PR #2, opened 2020-11-10
Update gh-pages URL path for MIDI and MP3 files

broadwell:
I'm not sure how long we'll continue using this particular gh-pages site for the MIDI files and MP3 files (and maybe other files, eventually), but `piano_rolls` just seemed like a more appropriate base path than `javatron`.


simonwiles (approved):

simonwiles:
Thanks Peter :)

PR #3, opened 2021-02-02
Switch to Snowpack

simonwiles:


broadwell:
Probably we should change
`    "build": "snowpack build",`
to
`    "build": "snowpack build --polyfill-node",`
in `package.json` as well -- it refuses to build otherwise.

Besides this, it looks good to merge.

simonwiles:
Yep, thanks.  There will certainly be other changes to the npm `build` script in the fullness of time (and it's a bit non-deterministic because `snowpack build` will work fine without `--polyfill node` if the cache has been primed, which I don't like at all), but this is adequate for now.

PR #5, opened 2021-04-07
Tempo

simonwiles:
Okay, let's get this rolling.  This is the simplest implementation of a tempo slider to start with.  Obviously this doesn't deal with the advanced functionality yet, which (I think?) is all connected with disabling the midi files' own timing instructions and/or replacing them?

Some notes for consideration, nonetheless:

w/r/t base tempo
1) will tempo changes always be available on the metadata track (i.e. track 0)?
2) would it be safe to simply look for an event with tick = 0?
3) in the event that we're generating our own metadata, `baseTempo` could perhaps be predetermined offline and then read from JSON to avoid any need to calculate at play time.

Also, there's an issue concerning the display for the tempo slider.  Three options that spring to mind:
1) try to show the actual bpm value (as computed from the roll instructions and the slider value);
2) show the value as a modifier (1x / 100% if unaltered or something);
3) don't show a value :)

Suggest we merge this and note these for addressing in the future, perhaps with an issue to return to or a note for discussion, as appropriate (unless it's blindingly obvious what the correct resolution is, in which case let's by all means resolve right away).

### DIFF
--- a/src/components/PlaybackControls.svelte
+++ b/src/components/PlaybackControls.svelte
@@ -5,7 +5,7 @@
 </style>
 
 <script>
-  import { pedalling, volume } from "../stores";
+  import { pedalling, volume, tempo } from "../stores";
 
   export let playPauseMidiFile;
   export let stopMidiFile;
@@ -66,4 +66,17 @@
     />
     {$volume.right}
   </div>
+  <div>
+    Tempo:
+    <input
+      type="range"
+      min="0"
+      max="180"
+      step="10"
+      bind:value={$tempo}
+      name="tempo"
+    />
+    {$tempo}
+  </div>
+
 </div>
--- a/src/components/PlaybackControls.svelte
+++ b/src/components/PlaybackControls.svelte
@@ -5,7 +5,7 @@
 </style>
 
 <script>
-  import { pedalling, volume, tempo } from "../stores";
+  import { pedalling, volume, tempoControl } from "../stores";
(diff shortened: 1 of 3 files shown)


broadwell:
Thanks! My takes on the issues from above:

I think it should be sufficient to get the base/starting tempo by always looking for a Tempo Change event at tick 0 of Track 0. Both seem to be the conventional locations for the starting tempo. So extracting this data into our own JSON metadata file is certainly feasible but wouldn't have much benefit beyond making our offline metadata a bit more comprehensive.

Not showing a tempo value at all actually seems to be the most "player-pianistic" option. The tempo slider controls on player pianos often seem to have had no markings at all, though on some models they would also move a pointer within an adjustable rain gauge-style indicator near the roll/tracker; as pianolists learned the inherent playback speed of their own piano's governor module, they could use this to get a ballpark numerical indication of how fast or slow it would be at a given setting. But the rolls themselves rarely if ever provided meaningful tempo information, which would have needed to be expressed as something like "quarter note = 1.5 inches at start of roll" -- hence the lack of this info in our metadata -- and the actual speeds of the players were not well standardized at all, so it's all pretty much anarchy and chaos.

broadwell:
And yeah, my expectation re: the advanced functionality is that it would probably involve the option to disregard the tempo instructions on the roll, compute our own tempo map based on specified parameters, and enact it at playback time.

PR #6, opened 2021-04-08
Progress

simonwiles:
This is a rudimentary progress slider for the basic MIDI playback.  The standout problem with it is that it only updates on MIDI ticks -- the solution to this will come when the roll viewer integration is added, I suppose.  Should merge cleanly to `main` once #5 is merged.

Also refactors the "stop → adjust → start again" logic to a reusable function -- not sure how useful this really is.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -4,6 +4,7 @@
     pianoReady,
     playPauseMidiFile,
     stopMidiFile,
+    skipToPercentage
   } from "./components/SamplePlayer";
   import RollDetails from "./components/RollDetails.svelte";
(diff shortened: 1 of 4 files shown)


broadwell:
Cool. If we plan to provide the option of hiding the roll viewer, then we'll probably want to have the progress slider/value update between MIDI events. A `setInterval()` handler run fairly often (like every 100 ms) can just do `midiSamplePlayer.getCurrentTick()`, and because MIDI ticks = pixels on the roll, the progress slider and roll position (if shown) can be updated in the same function.

simonwiles:
Thanks -- that's something I wanted to talk about face-to-virtual-face tomorrow :)

PR #8, opened 2021-04-08
Fancy slider

simonwiles:
This is a sort of initial scaffolding for future presentational components.  Range sliders, in particular, look so different across browsers that some custom styling is almost a requirement.  I've implemented mousewheel scrolling, although it seems in Firefox it only works when the input element has focus, which sucks a bit.  Haven't tested any of this in Safari, either.

The styling of the sliders is very tweakable (and can inherit from `globals.scss` for theming.  We may not want to use these as is, though, and very likely not for all controls.  For the progress bar, in particular, we might want something quite different.  There are also lots of ways these might be improved from a UI/UX point of view, that we can explore should it seem worthwhile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,9 @@
     "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-svelte3": "^2.7.3",
     "jest": "^26.6.1",
-    "node-sass": "^5.0.0",
     "prettier": "^2.1.2",
     "prettier-plugin-svelte": "^1.4.1",
+    "sass": "^1.32.8",
     "snowpack": "^3.0.10",
     "svelte-jester": "^1.1.5"
   },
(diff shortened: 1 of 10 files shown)


broadwell:
This does indeed look fancy, and I think it's a great foundation for what is likely to be one of the most heavily used UI components.

It looks fine on Chrome/Opera, Firefox, Safari for Mac. Scroll/mousewheel support works on all of those except Firefox. It would be great if the slider also supports left/right wheel/scroll events, as the Noh slider does, but we can work on that later.

simonwiles:
Thanks -- does scrolling work on FF if the input element has focus (e.g. if you click on it first)?  That's the behaviour I get here.

Re. left/right scrolling, I'll take a look at what we did for Noh -- something makes me think it was more complicated than simply testing the `deltaX` property of the mousewheel event, for some reason.

broadwell:
Sorry, just catching up on these replies now!

FF on Mac OS still doesn't seem to apply the scroll events to the slider even when it is in focus (has just been clicked).

I think this is the relevant code from Noh. Looks like the main trick is checking whether deltaX or deltaY is the "dominant" axis of scrolling: https://github.com/sul-cidr/noh/blob/66192279a7a804d4a92579057b54b41d2a66ce77/webpack/components/TimelineScrubber.jsx#L97

PR #9, opened 2021-04-09
Implement roll switching with a <RollSelector/> component

simonwiles:
Lots to be done here still, of course, but I needed `catalog.json` to get the iiif manifest urls, so this had to come first.  A couple of the MIDI files are missing because I couldn't find them on the SUPRA site (and didn't want to extract them from the file in your repo), so I took that as the nudge to do a bit of error-handling around the fetch failing (which I might have skipped for now otherwise).  This lead to a tiny notification component too.  It only has two styles right now -- the default (all white) and a slightly garish `error` style -- and isn't animated / modal / configurable / several other things that might or might not be desirable.


broadwell (commented):
Looks good! The missing MIDI files are for 88-note rolls from the Garcia-Sampedro collection, so they're not on the SUPRA site or repo, though they are in my repo and in the SDR. Anyway their absence provides a useful pretext for the error notification module.

PR #10, opened 2021-04-12
Active Notes

simonwiles:
Small PR to add an `activeNotes` store which wraps and exposes a `Set` object.


broadwell:
Great -- just what the, uh, data structure doctor ordered.

PR #11, opened 2021-04-12
Refactoring & housekeeping

simonwiles:
This is a bit of a of medley PR.

1) There is a bit of refactoring of the targets of the `playPause` and `stop` controls -- these should control the state of the app as a whole, which will then coordinate and delegate to components as appropriate.  Right now the app-level functions still call methods on the `midiSamplePlayer` directly, but this will probably wind up being abstracted.  `playbackProgress` is appropriately reset when the player is stopped, and (less clear cut) `activeNotes` is reset when the player is stopped or paused.

2) mp3 samples are added to and called from the repo.  I've only added the 175 files that are required to support `velocities: 2`, since that's what this version of the app is currently set at.  This will obviously bear some discussion.  There's probably a case for including all the samples anyway, even if they won't be called.

3) There's some linting and removing of old development comments and data.


broadwell:
Including the MP3 samples in the bundle is probably the way to go, though I'll just note here that loading them asynchronously is possible and would allow for 1) the app to provide some basic browse functionality while they're loading, and 2) dynamically changing the number of velocities used for the piano. But it's not obvious that either is a feature that we'll actually want in the released app. I do however think we'll probably want to be able to load the MIDI files asynchronously, assuming the app will cover more or less the full collection.

For the prototype, there were separate MIDI player instances for the roll and for the score (the latter intended to demo the score-following "bouncing ball" functionality independent of the roll view), but it makes sense to have just one MIDI player in this iteration of the app, and thus play/pause/stop apply to the entire app.

We may want to consider retaining the contents of `activeNotes` when the app is paused (but still clear them when it's stopped) -- though this eventually will involve tackling other more complicated interactions between the keyboard vis and the roll viewer and the progress scrubber. All in good time. In any case, we'll also need to add some logic that calls `piano.stopAll()` or runs `piano.keyUp()` on each active note on stop and maybe on pause, as requested by the PIs -- otherwise the strings continue to ring for quite a while.

simonwiles:
> Including the MP3 samples in the bundle is probably the way to go, though I'll just note here that loading them asynchronously is possible and would allow for 1) the app to provide some basic browse functionality while they're loading, and 2) dynamically changing the number of velocities used for the piano. But it's not obvious that either is a feature that we'll actually want in the released app. I do however think we'll probably want to be able to load the MIDI files asynchronously, assuming the app will cover more or less the full collection.

I'm not sure I can see any value in bundling the mp3 samples.  Did you have something in mind?  This PR does call them asynchronously, in case that wasn't clear (apologies if my prose was confusing!).

> For the prototype, there were separate MIDI player instances for the roll and for the score (the latter intended to demo the score-following "bouncing ball" functionality independent of the roll view), but it makes sense to have just one MIDI player in this iteration of the app, and thus play/pause/stop apply to the entire app.

Not 100% sure I'm following here either.  Are there circumstances under which we might want to have the roll image and a score progressing independently?

> We may want to consider retaining the contents of activeNotes when the app is paused (but still clear them when it's stopped) -- though this eventually will involve tackling other more complicated interactions between the keyboard vis and the roll viewer and the progress scrubber. All in good time.

Yes, not clear to me yet what the best behaviour is here, and it probably bears waiting until more of the pieces are in place.  There is a `roll-viewer` branch which is coming on, and bling-bling `keyboard` branch which I'm working on in parallel but keeping distinct for PR purposes (and factoring out the core functionality they require into separate smaller PRs as a I go along -- hence the `active-notes` PR and an imminent one to lay the groundwork for synchronizing roll-viewer and playback).

> In any case, we'll also need to add some logic that calls piano.stopAll() or runs piano.keyUp() on each active note on stop and maybe on pause, as requested by the PIs -- otherwise the strings continue to ring for quite a while.

I saw you'd added this to the prototype, but I've not looked at it yet.  Will add in due course :)

broadwell:
> I'm not sure I can see any value in bundling the mp3 samples. Did you have something in mind? This PR does call them asynchronously, in case that wasn't clear (apologies if my prose was confusing!).

Nope, not your fault -- I just got some wires crossed in my brain and was apparently equating "included in the build/deploy package" with "loaded synchronously." This way is fine, and including them in the package also means that anyone who forks the project will be more likely to load the files from their own server/account rather than pointing at the original repo's gh-pages branch (as is the case with ToneJS Piano).

> Not 100% sure I'm following here either. Are there circumstances under which we might want to have the roll image and a score progressing independently?

 No, almost surely not. Separating them in the prototype was purely for evaluation purposes.

PR #12, opened 2021-04-13
App timing control

simonwiles:
This is preparatory work that makes it trivial to sync automatic roll-viewer panning to playback.  It may prove to be a little simplistic, but we'll only complicate it further if actually necessary.

It will need some battle testing, but it will be appropriate to do that once the roll viewer is actually on the page :)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -5,6 +5,7 @@
     tempoControl,
     playbackProgress,
     activeNotes,
+    currentTick,
   } from "./stores";
   import {
     midiSamplePlayer,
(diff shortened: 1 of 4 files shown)


broadwell:
Looks good! Updating the progress percent value on each "playing" event from the MIDI player sure does make the numbers fly... Maybe this will be a suitable trigger for the roll scrolling animation (rather than an external timer) -- though the number of events may need to be throttled to avoid overwhelming the OSD viewer.

simonwiles:
Thanks yeah, I'm working on that now and I saw your note in the prototype code, but as yet I haven't been able to make it cause a problem (now zooming and panning while moving and playing, on the other hand...).

PR #13, opened 2021-04-14
Roll Viewer

simonwiles:
I had originally planned for this to be more feature-complete before PR-ing, but I think it's best I ask you to take a look at this now and we get the basic functionality in.  Panning the roll vertically and having the audio sync was giving me a headache, and I want to work on something else for a bit and come back to it :)  (I also have a thought about how that might be implemented differently, perhaps? better.)

Notes on this PR:

1) Zooming and horizontal panning should be rock solid -- if they're not in any environment for any reason, please let me know (can wait until merged and pushed to `gh-pages` for ease of testing on, e.g., mobile devices).

2) This implementation calls `panViewportToTick` on every tick (effectively from `midiSamplePlayer.on("play")`.  You have a note in the prototype code that reads:
    > syncing at every note effect also can cause problems on certain browsers if the playback events start to lag behind their scheduled times.

    I've not been able to make this cause a problem yet, but it wants some proper battle testing in different environments and on some crippled VMs and lower-powered devices.

3) This currently resets the roll on "stop".  Again you have a note about this, and in addition to your proposed callback solution we could also supply the `immediate` argument to the [`panTo`](https://openseadragon.github.io/docs/OpenSeadragon.Viewport.html#panTo) method.  Personally I think resetting on "stop" is the better behaviour (the progress slider does...), but I'll defer to you and the PIs :)

4) in any event there may be a case for scrolling `immediate`ly by default when the progress slider is moved, for example.  Something to consider (we could possibly toggle the behaviour at runtime with a discreet and/or hidden checkbox).

5) I have some ideas about styling and layout, but obviously they're not in here yet :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "@snowpack/plugin-sass": "^1.4.0",
     "@tonejs/piano": "^0.2.1",
     "midi-player-js": "^2.0.13",
+    "openseadragon": "^2.4.2",
     "svelte": "^3.29.4",
     "svelte-preprocess": "^4.7.0",
     "tone": "^14.7.58",
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -11,6 +11,7 @@
   import RollSelector from "./components/RollSelector.svelte";
   import RollDetails from "./components/RollDetails.svelte";
   import PlaybackControls from "./components/PlaybackControls.svelte";
+  import RollViewer from "./components/RollViewer.svelte";
   import Notification, { notify } from "./ui-components/Notification.svelte";
 
   const title = "Pianolatron Development";
@@ -96,5 +97,6 @@
 {#if appReady}
   <RollDetails />
   <PlaybackControls {playPauseApp} {stopApp} {skipToPercentage} />
+  <RollViewer imageUrl={currentRoll.image_url} />
 {/if}
 <Notification />
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -9,10 +9,30 @@
   import { onMount } from "svelte";
   import OpenSeadragon from "openseadragon";
 
+  import { rollMetadata } from "../stores";
+
   export let imageUrl;
+  let openSeadragon;
+
+  export const panViewportToTick = (tick) => {
+    if (!openSeadragon) return;
+    const { viewport } = openSeadragon;
+    const viewportBounds = viewport.getBounds();
+    const linePx =
+      parseInt($rollMetadata.FIRST_HOLE, 10) + (scrollDownwards ? tick : -tick);
+
+    const lineViewport = viewport.imageToViewportCoordinates(0, linePx);
+
+    const lineCenter = new OpenSeadragon.Point(
+      viewportBounds.width / 2,
+      lineViewport.y,
+    );
+
+    viewport.panTo(lineCenter);
+  };
 
   onMount(async () => {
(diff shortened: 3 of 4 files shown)


simonwiles:
Re: independently setting the transition speed for zooming as opposed to panning, it seems it is possible: https://stackoverflow.com/questions/25020629/can-i-slow-down-openseadragon-pan-zoom-animations

broadwell:
This all looks good to me. Definitely interested in hearing your ideas about alternative implementations. Re: your notes:

1. This works fine in all of my browsers so far. The horizontal pan-on-drag behavior is great. Let's talk sometime about whether vertical pan-on-drag should be supported during playback. A straightforward implementation is to pause the playback for the duration of a drag event, but this also causes playback to be interrupted during a horizontal pan, which often isn't desired.  
Relatedly, the PIs expressed interest in using manual zoom buttons, which are exposed via `showNavigationControl: true`, though the home and fullscreen buttons that accompany them seem much less useful and probably should be hidden in that case (note that all of the button images can be replaced/restyled, too). It seemed like the PIs might also appreciate a good old four-way pan control, but most image/map viewers eschew those nowadays.
2. I also notice no issues with this, though I agree we'll eventually be obligated to explore whether optimizations for underpowered devices may be needed and/or worthwhile.
3. I'm fine with resetting on "stop." I think the PIs were happy with this behavior, too, but they'll let us know if not.
4. OSD's default slippy zoom/pan behavior is also the default of the prototype, mostly due to laziness, but re: this note and also your comment about the possibility of modifying the zoom and/or pan speed -- we certainly can experiment with this; we might come up with something that matches the emerging gestalt of the app a bit better.

Re: why the prototype's tracker bar lags the highlighted notes: my best guess is that the tall viewport dimensions somehow result in the calculated center pixel of the viewable window (which should be the location of the current tick) not exactly lining up with the position that's 50% from the top of the div. But I haven't actually chased this one down, mostly because I was hoping the pseudo-element overlay would be replaced by a fancier SVG overlay feature eventually. Maybe that will still happen.

simonwiles:
Thanks Peter!  I'm not 100% sure I like the pan-to-scrub behaviour (jokes about turntable scratch noises aside...), but if you and/or the PIs want it I'm sure it can be done.  The reason it was causing me a headache was because of the way I'd set the roll viewer to auto-track the `$currentTick` store.  This is very fixable, just might require an extra state variable and I was contorting myself trying to avoid that eventuality (and then got fed up, temporarily :)).

I'm wondering about a mini-map -style control that would allow visual scrubbing (as well as giving a great overview of the roll, and possibly replacing the progress slider, which I'm not mad keen on either).  First glance suggests OpenSeadragon's built-in navigator might not cut it, but this idea was beginning to percolate before I knew that was even a thing, and I kinda expected to write it from scratch anyway.  When the mood is on me I might try a prototype for your consideration :)

Re. pretty much all of your other comments: thanks, let's roll with this (pun intended) for now and iterate as needed (especially w/r/t the tracker bar <-> highlighted notes misalignment -- re-implementing that will hit the top of my list real soon, and with any luck it will go away by itself :)).

broadwell:
Yeah, I also hear record-scratching noises in my head whenever I try to drag the roll during playback.

Using a zoomed-out (and probably rotated?) view of the roll as a progress indicator/scrubber is an intriguing idea. Agree that OSD's built-in navigator may not be up to the task, though.

PR #14, opened 2021-04-14
Keyboard

simonwiles:
This PR adds a hand-crafted keyboard display which is much more lightweight than the one from [`piano-keyboard`](https://www.npmjs.com/package/piano-keyboard) as well as being a bit blingier.

Notes:

1) For the "api" (I use the term loosely!) I've followed the module used in the prototype -- i.e., there's a `data-key` attribute on each key element with a key number in it, and a `title` attribute with the note name and octave.

2) There's a magic number `20` cribbed from the prototype and added in aff2b6a7.  Tbh I'm not 100% sure I've even got this right (figured it'd be obvious to you at a glance, so was too lazy to check properly myself given you'd need to okay it anyway :)).  If I understand/guess correctly, this is basically just because the key numbers in MIDI files start in a different place than the A0 key number for the `piano-keyboard` module?  If that's all it is, probably just hardcode the key numbers in this component to match the MIDI values.

3) The key-down effect styling is nice (imo, at least), but a bit subtle when the music's playing, so I added the blue highlight à la the prototype.  I'm sort of wondering whether to allow a toggle to enable/disable the blue -- but that's a part of a larger conversation, most likely.

4) Gecko (FF) and blink do different things w/r/t to the width of the console (the bit surrounding the actual keys -- dunno if that's an appropriate name), but that's probably better fixed later in a bit more context (work has begun).

5) The biggest "issue":  the keys are generated dynamically using svelte templating.  I'm not sure this is actually a very good idea.  There are pros and cons, but the main advantage is the ability to generate keyboards of different sizes.  However, even this functionality has significant limitations as implemented (width of keys isn't dynamic and can't be set at runtime, always begins from A0, etc.) and would need a bit of work to be really useful.  So -- are we likely to ever want/need this at all?  If not, I think I might be inclined to just write out the markup as a literal fragment, tbh.

6) You can't actually play it.  I see the prototype has this disabled in recent iterations.  Will add if requested :)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -99,6 +99,6 @@
   <RollDetails />
   <PlaybackControls {playPauseApp} {stopApp} {skipToPercentage} />
   <RollViewer imageUrl={currentRoll.image_url} />
-  <Keyboard keyCount="87" />
+  <Keyboard keyCount="87" {activeNotes} />
 {/if}
 <Notification />
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -99,6 +99,6 @@
   <RollDetails />
   <PlaybackControls {playPauseApp} {stopApp} {skipToPercentage} />
   <RollViewer imageUrl={currentRoll.image_url} />
-  <Keyboard />
+  <Keyboard keyCount="87" />
 {/if}
 <Notification />
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -12,6 +12,7 @@
   import RollDetails from "./components/RollDetails.svelte";
   import PlaybackControls from "./components/PlaybackControls.svelte";
   import RollViewer from "./components/RollViewer.svelte";
+  import Keyboard from "./components/Keyboard.svelte";
   import Notification, { notify } from "./ui-components/Notification.svelte";
 
   const title = "Pianolatron Development";
@@ -98,5 +99,6 @@
   <RollDetails />
   <PlaybackControls {playPauseApp} {stopApp} {skipToPercentage} />
   <RollViewer imageUrl={currentRoll.image_url} />
+  <Keyboard />
 {/if}
 <Notification />
--- a/src/components/Keyboard.svelte
+++ b/src/components/Keyboard.svelte
@@ -160,7 +160,9 @@
       .querySelectorAll(".depressed")
       .forEach((el) => el.classList.remove("depressed"));
     $activeNotes.forEach((note) =>
-      document.querySelector(`[data-key="${note}"]`).classList.add("depressed"),
+      document
+        .querySelector(`[data-key="${note}"]`)
+        ?.classList.add("depressed"),
     );
   }
 </script>
--- a/src/components/Keyboard.svelte
+++ b/src/components/Keyboard.svelte
@@ -78,7 +78,7 @@
         box-shadow: inset 0px -1px 2px rgba(255, 255, 255, 0.4),
           0 2px 3px rgba(0, 0, 0, 0.4);
         height: 120px;
-        left: -2 - ($key-width / 2 / 2);
+        left: calc(-2px - #{$key-width / 4});
         position: absolute;
         top: 0px;
         width: $key-width / 2;
--- a/src/components/Keyboard.svelte
+++ b/src/components/Keyboard.svelte
@@ -6,193 +6,128 @@
     border: 2px solid #333;
     box-shadow: 0 2px 0px #666, 0 3px 0px #555, 0 4px 0px #444, 0 6px 6px #000,
       inset 0 -1px 1px rgba(255, 255, 255, 0.5), inset 0 -4px 5px #000;
-    display: grid;
-    margin: 1em 0;
-    place-content: center;
+    margin: 1em auto;
     position: relative;
+    width: fit-content;
+    padding: 0 2em;
   }
 
-  ul {
+  div#keys {
     border-top: 2px solid #222;
     display: block;
     height: 240px;
     margin: 0;
     padding: 0;
     width: 100%;
-  }
 
-  ul li {
-    background: #aaa;
-    display: inline;
-    float: left;
-    list-style: none;
-    position: relative;
-    width: $key-width;
-
-    /* White Keys */
     div {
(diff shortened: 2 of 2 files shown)


simonwiles:
Don't know if you've had a chance to look at this yet, but I think I want to drop the "console" or whatever it is around the outside more-or-less altogether.  Think I should add that change to this PR, or you want to merge this as is and wait for a more general layout and styling PR that's incoming?

broadwell:
This is excellent! The keyboard used for the prototype was chosen just because it was easy to integrate; it wasn't even the best looking or the most lightweight among the available modules, but anyway I'm pretty sure this surpasses all of them in both respects.

1. (and 2.) The API is great. Using the magic number +20 is probably fine, or at least in keeping with the well-established convention for converting from piano key numbers to MIDI note numbers (piano 1 = piano A0 = MIDI 21). A slightly cleaner option would be to number the keys starting at 21 and to make sure that the rest of the app also refers to keys by their MIDI note numbers. The ToneJS Piano component already does this. In any case, using numbers, either MIDI or piano key, is far preferable in a programming context to using the names because the black keys all have two identities (although I think the "sharp" names are used more commonly). Sometimes, though, it's helpful for debugging/interpretability purposes to keep the names around, e.g., the ToneJS piano sample files are named this way.

5. Having the keyboard/key sizes be responsive is already a huge improvement over the prototype. I can't think of a situation when we'd ever want to show less than the full set of keys, but there is a possible application of being able to set the key widths dynamically, which is that with a lot of fiddling (and assuming accurate hole/column with metadata for the roll), they could be aligned with the roll at a set zoom level to produce the popular "piano Tetris" effect: https://www.youtube.com/watch?v=LjVY3ftT6VI. This is obviously not at all representative of the actual experience of playing a piano roll, so it's unlikely the PIs will want it, but it also looks cool, so who knows... maybe they will?

6. The prototype's keyboard isn't, um, intentionally made unplayable -- it's actually still playable until the roll is displayed, at which time it for some reason becomes deaf to click events unless the window size is changed. In any case, I think it's generally desirable for the keyboard to be playable; pianolists seem not to play duets with the rolls very often, but it's always an option with a real player piano, so having a "mute" keyboard would be overall a net minus for the look and feel of the app, I think. 

Re: 3-4 and the update: I'm on the fence about removing the surrounding housing/console (the front bit is the "keyslip" and the ones to the sides are the "keyblocks"; there doesn't seem to be an official collective term for them). Finding the right level of skeumorphism likely will require some ongoing and possibly tedious negotiation, but regardless, my two suggestions would be:
- Maybe add a thin horizontal red strip just above the keys, to represent the bit of felt that sticks out there.
- I think it would be preferable for the gap that appears above/at the base of a white key when it is pressed to be black/dark (suggesting shadow) rather than white/clear.

simonwiles:
+1 for "Apocalypse Mode"

broadwell:
Also re: making the piano playable, it would be nice if it were possible to play a slide glissando by clicking on one key and then dragging sideways over several others. Probably this would involve adding/modifying some mouse click/drag event handlers?
Just making a note of it here; these are rather low-priority features given that a point-and-click or touch interface is only ever likely to let you play the virtual keyboard with one finger or maybe two, "Chopsticks"-style. MIDI device input support is by far the best way to accommodate true playability, I think.

simonwiles:
Thanks yeah -- it had occurred to me too that this would be nice (glad to have the appropriate terminology at my disposal now, though :)).  It's probably an order of magnitude more difficult than simple `on:click`s, but from a low base; should be easy enough, I think.

simonwiles:
> I think it would be preferable for the gap that appears above/at the base of a white key when it is pressed to be black/dark (suggesting shadow) rather than white/clear.

Question: is this addressed at all by the felt strip being in place, and the shadowing that comes with that?  If not, I might need you to explain what you have in mind in more detail.

src/App.svelte
```diff
@@ -98,5 +99,6 @@
   <RollDetails />
   <PlaybackControls {playPauseApp} {stopApp} {skipToPercentage} />
   <RollViewer imageUrl={currentRoll.image_url} />
+  <Keyboard keyCount="87" {activeNotes} />
```
broadwell, line 102: `keyCount` should be 88 (the top key, C8, is missing).

src/components/Keyboard.svelte
```diff
@@ -0,0 +1,161 @@
...
+      }
+      :global(:nth-child(2).depressed) {
+        background: $active-key-highlight;
+      }
+    }
+  }
+</style>
+
+<script>
+  export let keyCount = 87;
```
broadwell, line 103: Also should default to 88.

broadwell:
Yeah the red felt strip solves my concern about the top of the white keys looking weird when pressed. I do however feel that the red strip could stand to be quite a bit thinner, but we can finesse this later.

simonwiles:
Got it -- wasn't sure what I was picturing.  Wanna play with the dev. tools and suggest a value?  I'll correct the `keyCount` (and perhaps change this too) in an upcoming grab-bag of miscellaneous fixes.

broadwell:
Sure, after a few minutes of leaning over my keyboard and squinting, I think these are the settings I like the best:
```
    position: absolute;
    top: 11px;
    height: 2px;
    left: 0px;
    right: 0px;
    box-shadow: 0 1px 1px rgb(0 0 0 / 70%), 1px 0px 1px rgb(0 0 0 / 70%), -1px 0px 1px rgb(0 0 0 / 70%);
```

simonwiles:
That thin?  Okay cool!  (I was miles out).  Adjustment incoming!  (I'd also already made it so the felt colour follows `$primary-accent`, 'cos I thought that looked cool.  It's not gonna be that noticeable now, but I might as well leave it in anyway, for now at least.)

src/components/Keyboard.svelte
simonwiles: Addressed by 15a40e7b.

src/App.svelte
simonwiles: Addressed by 15a40e7b.

PR #15, opened 2021-04-15
Layout rearrangement

simonwiles:
This is a start on rearranging the components into something resembling an interface.  See what you think.  It's all provisional and up-for-grabs, of course.

1) When the background colour is anything other that white, the white non-transparent background on the roll images is a bit obnoxious, so there's a solution here that uses CSS blend modes to make the white bits match the chosen `$background-color`.  There are some downsides to this, and it will only work for flat background colours, but I sorta feel like it's this (or something very like it) or a plain white background, tbh -- interested to know what you think.  I do have a commit on the [`manual-blend`](https://github.com/sul-cidr/pianolatron/tree/manual-blend) branch which does basically the same thing but calculates the needed overlay value at build-time with a binary search in an SCSS function, and hard-codes the value.

2) This PR includes a felt strip over the top of the keys.

3) The pedal-related buttons should probably go beneath the keyboard.

4) etc...

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -15,8 +15,6 @@
   import Keyboard from "./components/Keyboard.svelte";
   import Notification, { notify } from "./ui-components/Notification.svelte";
 
-  const title = "Pianolatron Development";
-
   let appReady = false;
   let mididataReady;
   let currentRoll;
@@ -93,7 +91,6 @@
   $: playbackProgress.update(() => $currentTick / midiSamplePlayer.totalTicks);
 </script>
 
-<h1>{title}</h1>
 <RollSelector bind:currentRoll />
 {#if appReady}
   <RollDetails />
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -1,3 +1,31 @@
+<style>
+  #app {
+    display: grid;
+    grid-template-rows: 1fr auto;
+    grid-template-columns: auto 1fr auto;
+    grid-template-areas:
+      "left roll right"
+      "keyboard keyboard keyboard";
+    height: 100vh;
+  }
+
+  #roll-details {
+    grid-area: left;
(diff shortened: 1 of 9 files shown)


broadwell:
This is a great first iteration. A few comments:
- I like the three main columns, and feel that giving the user the option to swap out the contents of certain columns with others might be worth exploring; for example, the metadata could be temporarily replaced by a score view if the user desires (and if available), or even by a visualization of the expression parameters. 
- Responsiveness could be improved, but there's probably not much point in spending a lot of time on that until we have a better handle on the full set of components and more of a consensus on where they should go.
- The CSS blend mode is a really promising way of providing a non-white background. We'll have to see how this particular color looks with other roll types; the paper of the 88-note rolls from the Garcia-Sampedro collection is a similar hue, which might make the holes difficult to see -- perhaps we could switch background colors depending on roll type?
- I agree the pedal controls should go beneath the keyboard but I like the idea of putting the other "pianolistic" controls (volumes, tempo, play/stop, eventually the "accent" button) in the upper right. It might be more true-to-life to put them below/in front of the keyboard, but as we see with the prototype's layout, this results in too many controls being way at the bottom of the screen.

simonwiles:
Thanks -- the background overlay filtering works pretty well with the 88-note roles imo (PR that adds the missing MIDI files incoming), but I've just been playing with inset box shadowing on the roll-viewer div and it does a reasonable job of creating a kind of window for the roll images to scroll in that doesn't make the white edges look too bad.  Not sure, tbh.

Responsiveness is on my list -- this PR uses a css grid layout, so it's pretty easy to move things around at runtime or with media queries or whatever, so as it become clearer what we might want, it should be relatively straightforward to implement it without re-architecting much, if anything.

PR #18, opened 2021-04-19
Loading throbber

simonwiles:
This is a little unpolished, but it's a needed addition, and will be plenty adequate for now, I think?


broadwell (commented):
Yes some loading interstitial indicator is certainly needed. I'm sorry to say, though, that I can't actually see the loading message(s) or throbber effect here, even with the network throttled to "slow 3G" speeds on Opera and Firefox. Is something missing?

simonwiles:
It's working okay for me here (FF 88 and Opera 75, both on Linux of course).

There's no auto-prefixing on the CSS (for now), so my first thought is that if your FF is a bit old, there may be something I've used that's not supported in an unprefixed way -- but I've had a look and it seems to be all old and very widely-supported syntax, so I'm not sure.  I've done a deploy of the throbber code to https://sul-cidr.github.io/pianolatron/ -- could you check that and let me know if it's still missing for you?

I'll start firing up some VMs :)

PR #20, opened 2021-04-20
Accent button

simonwiles:
Mechanics of the "accent" button.  Interested in ideas for styling and placement -- I have a couple, but I'm not confident enough in them to even implement them as a conversation starter.

### DIFF
--- a/src/components/PlaybackControls.svelte
+++ b/src/components/PlaybackControls.svelte
@@ -133,7 +133,8 @@
     class:pedal-on={$pedalling.accent}
     aria-pressed={$pedalling.accent}
     on:mousedown={() => pedalling.update((val) => ({ ...val, accent: true }))}
-    on:mouseup={() => pedalling.update((val) => ({ ...val, accent: false }))}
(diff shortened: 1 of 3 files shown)


broadwell:
Regarding positioning: I was imagining that the accent button might remain with the top-right cluster of controls even after the soft and sustain pedals eventually migrate below the keyboard. Technically, accents were sometimes effected via the pneumatic foot pedals, but that was kind of a different class of control from the regular piano pedals, and anyway I think accent/sforzando buttons also were a thing, so keeping it near the other controls that a pianolist would manipulate by hand (tempo, volume sliders) makes some intuitive sense to me. It might however be located some distance from the stop and play buttons, which aren't really in the same class of performance-altering controls.

Styling-wise, I kind of like the idea of the accent button having a different shape than the start/stop buttons -- perhaps it could be about twice as wide as they are -- and maybe even lighting up with a different color while pressed. This is the extent of my musings on the topic so far, though.

simonwiles:
Thanks!  Yeah, I was starting to wonder about something round, but then I realized I was thinking about something that you could hold down to increase the effect, and that's not really the right behaviour here, is it?

PR #21, opened 2021-04-21
Read json metadata

simonwiles:
This PR begins the process of drawing on (an) additional file(s) beyond the `.mid` to generate the interface.  As such it also opens the conversation about exactly how to do this :)

For the purposes of getting started, the JSON files have been committed to this repo, but not the script which generates/assembles them, which is presently in a personal repo (the files in this commit were created from this version:  https://github.com/simonwiles/pianolatron-data/blob/1cf25ed677ff24b1cd0bb0b877c9e785dc7c1215/build-metadata.py).  We need to decide how much of this apparatus belongs here, how much in a separate (presumably sul-cidr) repo, and how much should be considered transient artifacts that aren't persisted (e.g. the MODS, midi, and .txt files that are needed to generate the JSON).

There are also problems with some of the metadata itself, of course.  Some things related to the MODS files that stand out and that are directly related to this PR:

1) The MODS files for all the rolls in our little development corpus are converted from MARCXML, so something different may be necessary for the "born-MODS" stuff (will try to find an example, or ask Kevin);

2) The metadata presented in the MODS files is different (sometimes inferior?) to what's in the MIDI metadata tracks (where the latter exists).  Not sure exactly how this works given that the latter was created on the basis of the former?  Depending on what we actually want to have in the Pianolatron app interface, some investigation may be required here.

3) The MODS for the 88-note rolls don't appear to contain anything corresponding to a performer/instrumentalist.  Is this perhaps to be expected, given the nature of the non-reproducing rolls?

4) Some of the metadata may need some post-processing (see f84b7b8 in this PR and https://github.com/sul-cidr/pianolatron/blob/bc9a3306e7cbae30b741cd4f0b0addda75c6a7a5/src/assets/json/pz594dj8436.json).

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -34,6 +34,7 @@
     playbackProgress,
     activeNotes,
     currentTick,
+    rollMetadata,
   } from "./stores";
   import { midiSamplePlayer, pianoReady } from "./components/SamplePlayer";
   import RollSelector from "./components/RollSelector.svelte";
@@ -48,7 +49,6 @@
   let metadataReady;
   let currentRoll;
   let previousRoll;
-  let currentMetadata;
 
   const playPauseApp = () => {
     if (midiSamplePlayer.isPlaying()) {
@@ -110,7 +110,10 @@
         if (metadataResponse.status === 200) return metadataResponse.json();
         throw new Error("Error fetching metadata file! (Operation cancelled)");
       })
-      .then((metadataJson) => (currentMetadata = metadataJson))
+      .then(
+        (metadataJson) =>
+          ($rollMetadata = { ...$rollMetadata, ...metadataJson }),
+      )
       .catch((err) => {
         notify({ title: "Error!", message: err, type: "error" });
         currentRoll = previousRoll;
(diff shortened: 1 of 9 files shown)


broadwell:
This is a good start. More conversations about how to manage and possibly augment the metadata are sure to follow. Some comments:
2. My hunch here is that the metadata in the MIDI files sometimes was manually corrected, possibly by Craig. Not sure if there's a record of this in the relevant Github repos -- will check.
3. Yes, 88-note rolls normally won't have performers because they tended to be constructed/programmed rather than recorded. We can ask the PIs whether it's preferable just to hide the performer field for any 88-note roll -- we should be able to determine roll type pretty easily from our combined metadata, even if it's not easy to extract from the MODS (it's just in an arbitrary `<note>` element).
4. This record in general is quite vexing; the `<titleInfo>` field was obviously parsed, quite badly, from the `<objectLabel>` contents in the Searchworks record. Using the "uniform" flavor of the `<titleInfo>` would be preferable for this particular record, but what about the others?

PR #22, opened 2021-04-21
Hole highlighting v1

simonwiles:
So this is hole-highlighting v1.  The JSON files have been updated with the needed data, which swells them considerably -- I've already brought that down a tonne from the first go at 4b98f02, but it can come down further by remappng key names (to `x`, `y`, `w`, `h`, etc.), and that's probably worth doing at some point.

The major difference between the implementation here and the one on the prototype (aesthetic flourishes aside) is the use of an `IntervalTree` to keep track of what notes are on when.  The events exposed by the midiplayer module skip ticks, and so a simple structure that maps ticks to holes that start at that time isn't great.  It works better on the prototype than here because the prototype updates the hole highlighting on `midiEvent`, which gets the notes, but misses some control holes that don't trigger events (this may or may not be desirable, I suppose).  Since I've been using the `playing` event here, a map wasn't cutting it at all.

There are some unfinished parts here, partly because I've moved on for now to an alternative implementation that uses a single SVG overlay for the whole roll.  This has many advantages but also a few disadvantages, but it builds on this so I'd rather get this in first.


src/components/RollViewer.svelte
```diff
@@ -43,10 +88,42 @@
   import { rollMetadata, currentTick } from "../stores";
 
   export let imageUrl;
+  export let holesByTickInterval;
+
+  const WELTE_MIDI_START = 14;
+  const WELTE_RED_FIRST_NOTE = 28;
+  const WELTE_RED_LAST_NOTE = 103;
```
broadwell, line 95: So... turns out that the Feuerzauber roll goes up to G7, so this should be at least 107. Annoyingly, this conflicts with the comments in Craig's code, which suggest that the range should be MIDI 24-103 inclusive for red Welte rolls (yet the generated MIDI still plays correctly, suggesting that some part of the coce uses a different range).
https://github.com/pianoroll/roll-image-parser/blob/131bf54ab27803618c92093365dd2986a84d3d8b/src/RollOptions.cpp#L281
We'll need to get to the bottom of this eventually...

simonwiles: Oh joy...  If (*if*) this is the only place we'll need these constants (i.e. for displaying identifying information on hole overlays) then I'd be inclined to add a `type` (`control|note` or whatever) and a `note` key to the hole data, and then we can do whatever awkward shenanigans we need to do at build time.

simonwiles: Am I misreading Craig's notes or yours?  -- doesn't he say that G7 == 103?  where does 107 come from?

broadwell: Further perusal of the other rolls (especially the Liszt, which has a wide range and also seems to employ every control hole) suggests that these values may be our best bet, pending more authoritative input:
```
WELTE_RED_FIRST_NOTE = 28
WELTE_RED_LAST_NOTE = 107
```

broadwell:
This is a big improvement over the prototype in both implementation and aesthetics.

Being able to access the overlays for notes/holes that aren't "active" does seem like it might be preferable, though.

Also, I wonder if an option that displays the overlays (including note names) or some slightly more minimalist version of them when each note is played might be what the PIs have in mind. Whether such a viz is feasible is a different question of course -- avoiding overlap seems likely to be a challenge.

src/components/RollViewer.svelte
broadwell: OK yeah, Craig's numbers are the right ones. I was confused due to the MIDI numbers for the overlays and hence also the note names on the overlays being to be too high by 4 steps (a major third). The giveaway is that they don't actually match up with what the keyboard is showing (i.e., the actual MIDI note events). I'm tracking down the error now...

broadwell: So a quick fix for the overlay names (and numbers) is to use the following constants, but we definitely should review these values and how we use them -- labeling holes with their note names or control functions ahead of time in the JSON file may very well be the best path.
```
  const WELTE_MIDI_START = 10;
  const WELTE_RED_FIRST_NOTE = 24;
  const WELTE_RED_LAST_NOTE = 103;
```

simonwiles: Okay, great -- I'll commit that update directly to `main` so I can do a `gh-deploy` this morning.  Thanks!

simonwiles:
Cool, thanks for merging.  I've had another idea so I might rework the SVG-based v2 again to give it a try (not sure if it will actually work -- I really want the best of both worlds, but there are some obstacles in terms of the advantages and disadvantages using an SVG brings).

PR #24, opened 2021-04-24
Double overlays

simonwiles:
Using HTML-based overlays for all holes absolutely tanks performance, as you discovered with the prototype, and I couldn't find a way to do anything about that (OSD recalculates the positioning and size of every element when the viewer pans -- I'm not sure it really needs to tbh, but I didn't fancy delving far enough in to it right now to try to improve the behaviour, much less put together something that could be PR'd upstream).

A single SVG overlay works great, because OSD is only aware of the top level element, and the browser/SVG takes care of the rest.  The problem is that styling options are very limited with the SVG route (can't apply `mix-blend-mode` or similar differentially on individual elements within the SVG; dimensions are scaled with the SVG, rather than staying the same size through zoom events; animations and [pseudo-]pseudo-elements can be done, but I'd have needed to start from scratch with those), and I quite liked what I already had.

Eventually it occurred to me that I could have the best of both worlds if I just literally did both.  This PR adds an SVG overlay to the entire roll *and* uses the HTML overlay for active and hovered holes.  There's a tiny bit of glitchy-ness I had to work around concerning `mouseout` events firing inconsistently, but otherwise this is pretty tight, I think.

Outstanding:
* More information/formatting/whatever for the popups.

* Differential colouring for the control holes (desirable, presumably, but exactly what/how is not completely clear to me yet).

* There's a branch called `popup-when-active` with a one-line change on top of this PR that activates the "popups" (for want of a better expression) when the notes are active.  The styling is as here (the heavy `darkturquoise` border), so it doesn't look great, but it's something to look at.  I suppose this is the kind of feature that we would want to be toggle-able from a settings panel of some kind, so I might start working on one of those, soon, so we've somewhere to put UI that we don't want on the main screen.  Ideas for what this might look like (modal window?  slide-out drawer?  accordion of some sort?) very welcome!

* As I write this it occurs to me that since the SVG does nothing in this implementation except provide targets for `mouseover` events, with a bit of simple maths the targets could be made larger to make them easier to hit (it can be a bit fiddly to target small holes, especially at low zoom levels).  I'm not certain it would work, but as I think about it now I reckon it could be okay.  Your thoughts welcome!

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -58,6 +58,7 @@
   let currentRoll;
   let previousRoll;
   let holesByTickInterval = new IntervalTree();
+  let holeData;
 
   const buildHolesIntervalTree = (holeData) => {
     const scrollDownwards = $rollMetadata.ROLL_TYPE === "welte-red";
@@ -103,6 +104,7 @@
     tempoControl.set(60);
     volume.update((val) => ({ ...val, left: 1, right: 1 }));
     holesByTickInterval = new IntervalTree();
+    holeData = null;
   };
 
   const skipToTick = (tick) => {
@@ -148,8 +150,10 @@
     Promise.all([mididataReady, metadataReady, pianoReady]).then(
       ({ 1: metadataJson }) => {
         $rollMetadata = { ...$rollMetadata, ...metadataJson };
-        if (metadataJson.holeData)
-          buildHolesIntervalTree(metadataJson.holeData);
+        if (metadataJson.holeData) {
+          holeData = metadataJson.holeData;
+          buildHolesIntervalTree();
+        }
         appReady = true;
         previousRoll = currentRoll;
       },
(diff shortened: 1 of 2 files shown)


broadwell:
Truly the best of both worlds! I've had an inkling in the back of my mind for a while that it would be great and probably necessary to have two classes of overlays, one "background"-level overlay for all of the holes and a different one for the "active" holes, but this makes it a reality in a clean and non-performance-tanking way.

I do think the `popup-when-active` behavior is worth discussing the with the PIs. Placement of the active hole info overlays seems potentially challenging, but some overlap is probably fine, and a version of this in which the overlays fade out (while continuing to scroll) after the hole is no longer active might look really cool.

Re: the settings panel, I've kind of been thinking of it as sliding in from the side, and perhaps temporarily covering the right-hand controls panel or else the left-hand metadata/chooser panel. The visual metaphor I've had in mind is something like the side ranks of stops on an [organ console](https://en.wikipedia.org/wiki/Organ_console) (the things to the left and right of the keyboards), but like, you know, movable. This may be a terrible idea though.

simonwiles:
Thanks!

> I do think the popup-when-active behavior is worth discussing the with the PIs. Placement of the active hole info overlays seems potentially challenging, but some overlap is probably fine, and a version of this in which the overlays fade out (while continuing to scroll) after the hole is no longer active might look really cool.

Cool, I'll take another look before the next PI meeting.  Depends on how much information we want to put in the popup, of course, but for a display of that kind my mind jumps to having the note name appear over the top of the hole in a centered way, just white text with a heavy shadow for readability.  In my head, at least, I can't see this presenting any challenges.

PR #25, opened 2021-04-27
Roll viewer controls

simonwiles:
This is a first pass at some roll viewer controls.  The ones OpenSeadragon provides aren't really what we want, and the options for customizing them are such that we are better off just doing our own.

This PR includes basic functionality -- I'm expecting a subsequent PR to address:
1) making the fit-to-width and center buttons work when the roll is playing;
2) add automatic `disabled` styling to indicate when max/min zoom has been reached, and when the roll is already centered / at fit-to-width;
3) probably refactor into a separate Svelte component;
4) other things?

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -22,6 +22,7 @@
 
   #roll {
     grid-area: roll;
+    position: relative;
   }
 
   #audio-controls {
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -301,7 +301,11 @@
 
 <div id="roll-viewer" />
 <div id="roll-viewer-controls">
-  <button>
+  <button
+    on:click={() => {
+      openSeadragon.viewport.zoomBy(1.1);
+    }}
+  >
     <svg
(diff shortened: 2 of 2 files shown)


broadwell:
Cool, these look good to me! A slightly bigger zoom in/out delta might be preferred, like +/- .2.  
Also, I think the following icon designs or something like them might be a bit more intuitive (though maybe only to me):

zoom to fit (horizontally):
zoom-to-fit
center:
center
but I also appreciate the minimalism of the current icons, and think people will clue in to what they do very quickly regardless.

In any case, we can wait to see what the PIs have to say about all of this before making any changes.

simonwiles:
Thanks -- yeah, I was thinking about a bigger zoom delta too.  The OpenSeadragon docs say that `1.1`/`0.9` is the default, but whatever the jumps are when you mousewheel is slightly bigger than that, and feels much better.  I reckon the centering function is pretty useless, and it complicates something else I've been trying to do, so I've more-or-less made my mind up to just drop it at this point, unless you feel strongly.

PR #26, opened 2021-04-28
Roll viewer controls pt. 2

simonwiles:
Here are a couple more updates for the roll viewer controls, then.  The user-facing changes are:

1) the buttons are now disabled when their use isn't appropriate (imo this is quite important for decent UX);

2) the fit-to-width icon is changed a little.  I'm not sure anything much more detailed is going to work well at this size, so I did this quickly.  Willing to revise again as needed;

3) you can now fit-to-width successfully while the roll is playing (this will have to be revisited, I expect, in order to make scrubbing and/or vertical panning possible, but that's for later);

4) I dropped the 'center' button -- it poses a number of technical and UI issues, and the functionality isn't actually very useful imo; again, willing to reconsider.

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -93,7 +93,6 @@
 
 <script>
   import { onMount } from "svelte";
-  import { fade } from "svelte/transition";
   import OpenSeadragon from "openseadragon";
   import { rollMetadata, currentTick } from "../stores";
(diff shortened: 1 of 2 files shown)


broadwell:
Cool, I think the "fit to width" button provides all the centering functionality anyone would want. The current icon is fine, especially given that there's no need to distinguish it visually from a "center" button.

Functionality while playing is great, as is disabling buttons when their use isn't appropriate.

PR #27, opened 2021-05-03
Collapsible Sidebars

simonwiles:
This is still very much a work-in-progress, and it is quite likely that very little of this will survive in the codebase for very long, but since it seemed to meet with provisional approval today I figured I'd PR it now as a step towards reconciling the various branches and approaches I've been taking to reorganizing the layout.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -12,9 +12,9 @@
   #roll-details {
     grid-area: left;
     max-width: calc(348px + 2em);
+    padding: 1em;
 
     p {
-      margin: 1em;
       opacity: 0.5;
       padding: 0.5em 1em;
     }
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -1,21 +1,19 @@
 <style lang="scss">
   #app {
-    display: grid;
-    grid-template-rows: 1fr auto;
-    grid-template-columns: minmax(300px, 400px) minmax(800px, 1fr) minmax(
-        300px,
-        auto
-      );
-    grid-template-areas:
-      "left roll right"
-      "keyboard keyboard keyboard";
     height: 100vh;
+    display: flex;
+    flex-direction: column;
+    overflow: hidden;
+
+    > div:first-child {
+      display: flex;
+      flex: 1 0 auto;
+      position: relative;
+    }
   }
 
-  #roll-details {
-    grid-area: left;
-    padding: 1em;
-
+  :global(#roll-details) {
+    padding: 0.5em;
     p {
       opacity: 0.5;
       padding: 0.5em 1em;
@@ -23,16 +21,15 @@
   }
 
   #roll {
-    grid-area: roll;
     position: relative;
+    flex: 1 0 auto;
   }
 
   #audio-controls {
-    grid-area: right;
   }
 
   #keyboard-container {
-    grid-area: keyboard;
+    flex: 0 1 auto;
   }
 </style>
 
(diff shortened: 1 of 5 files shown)


broadwell:
Yes I agree this is worth merging as a basis for further layout enhancements.

FYI, that annoying auto-zoom behavior of the OSD viewer on sidebar expand/collapse can be disabled by specifying
`preserveImageSizeOnResize: true` when initializing the viewer.

Also, unrelated to this PR, I'm getting a `<RollViewerControls> was created without expected prop 'strafing'` warning whenever the controls appear, presumably related to the use of `bind:` when the `<RollViewerControls>` component is instantiated at the end of RollViewer.svelte. I'm honestly still rather confused by how binding to component props seems to work vs. how the docs say it's actually supposed to work. Maybe it would be preferable just to keep `strafing` in a store if it's really an issue?

simonwiles:
> Yes I agree this is worth merging as a basis for further layout enhancements.

Thanks, yeah, there's plenty more to come here, but this is an incremental improvement even if it doesn't reflect the end goal terribly well yet :)

> FYI, that annoying auto-zoom behavior of the OSD viewer on sidebar expand/collapse can be disabled by specifying
> `preserveImageSizeOnResize: true` when initializing the viewer.

Yeah, the problem with that is that then the image doesn't expand to use increased available space (or zoom out so you can still see the same portion when the available space gets smaller).  I spend some time trying to do it in app code (see my "savepoint" at [`osd-resize-bug`](https://github.com/sul-cidr/pianolatron/tree/osd-resize-bug)), but I wasn't able to make it work to my satisfaction without it going haywire if you tried it when the roll was actually playing.  I call it a bug with OSD because the image auto-resizes when the space gets smaller, but not when it gets bigger, and I was summoning the gumption to see if I could fix that in the OSD code.  However, revisiting it briefly now, `preserveImageSizeOnResize: true` is clearly a better behaviour than not, so I'll add that for now (thanks!), and make a note to maybe come back to this (either with a `ResizeObserver` and some custom code or take a run at modifying the OSD behaviour).

> Also, unrelated to this PR, I'm getting a `<RollViewerControls> was created without expected prop 'strafing'` warning whenever the controls appear, presumably related to the use of `bind:` when the `<RollViewerControls>` component is instantiated at the end of RollViewer.svelte. I'm honestly still rather confused by how binding to component props seems to work vs. how the docs say it's actually supposed to work. Maybe it would be preferable just to keep `strafing` in a store if it's really an issue?

Yeah, these pop up from time to time -- thanks for mentioning it.  They're dev.-only warnings, and are generated when the a prop is `undefined` at component-creation time.  They don't cause issues unless they do, if you see what I mean, but should be addressed.  In this case it's perfectly safe (`undefined` is falsey, which is the initial state of `strafing`), but the warning can be silenced simply by giving `strafing` an explicit initial value (i.e. `let strafing = false`) -- I'll do that.  (Apologies for the torturous variable name, btw -- it's intended to indicate that the role is panning with a horizontal component, whether that be from a dragging action or a re-centering action.  Better ideas welcome!)

broadwell:
Oh also re: alternative variable names for `strafing` -- this is indeed tricky because many of the others that come to mind are either cumbersome or else collide semantically with other related concepts (e.g., "panning"). "translating" (side-to-side motion in aviation) is about the best I can come up with, though again there's conceptual collision. Otherwise maybe "sidling", "flanking" (another martial term), "lateraling", "crossPanning", "kenBurnsing"?

PR #28, opened 2021-05-04
Misc Updates

simonwiles:
Another collection of small bits and pieces from the last week or so.

1) A "dowloading roll image..." note.  This one's been sitting ready to PR for ages -- I think it's needed.  The `fully-loaded-change` event on the image set seems to be the best available option.  It's not perfect, but nothing better seems to be exposed by OSD.

2) A quick footer.  I found the "Powered by CIDR" tag line on a mothballed Kindred London server, and decided it deserved to be resurrected (provisionally, at least) :)  The text is perhaps a bit dark, but you could say that about the whole site -- improving the body text is on the roadmap.

3) Some colour theming (I personally rather like the green, but it is a bit similar to Noh).  More to come here.

4) Stop notes from ringing on when playback is stopped / paused.

5) Prevent dev. warnings about passing uninitialized props.

6) Set `preserveImageSizeOnResize: true` on the OSD viewer -- this is certainly better than `false`.  I may try to do better with this at some point in the future.  Or maybe this is just fine.


broadwell:
I like the footer. "Powered by CIDR" has potential, I think. The updated color scheme also seems appropriate. Noh theaters, concert halls, and pianos are all pretty heavy on wood colors and other natural tones, so some commonalities are likely.

A couple of items for possible future updates:
- There seemed to be widespread support for changing the "Stop" button to a "Rewind" button.
- Is it supposed to be possible to advance or back up the playback progress by dragging the roll image up and down while stopped/paused? Currently it just returns to the previous playback point upon play resumption; the "Progress" scrollbar therefore is the only way to scrub back and forth. I believe we agreed that allowing scrubbing via roll-dragging *during* playback wasn't desirable (scratching vinyl and all that), but doing so while paused is probably OK?

simonwiles:
Thanks!

> The updated color scheme also seems appropriate. Noh theaters, concert halls, and pianos are all pretty heavy on wood colors and other natural tones, so some commonalities are likely.

I've got a settings panel ready for initial PR, and for want of anything better to put on it immediately, there's a theme switcher...

> There seemed to be widespread support for changing the "Stop" button to a "Rewind" button.

Agreed --  it's on my list, but there are a number of other changes I want to make to the playback buttons (including separating them, logically at least, from the pedals), and I'll wrap this change up with those.

> Is it supposed to be possible to advance or back up the playback progress by dragging the roll image up and down while stopped/paused? Currently it just returns to the previous playback point upon play resumption; the "Progress" scrollbar therefore is the only way to scrub back and forth. I believe we agreed that allowing scrubbing via roll-dragging during playback wasn't desirable (scratching vinyl and all that), but doing so while paused is probably OK?

Yeah, this definitely needs doing -- the lack of this functionality has been annoying me too.  I suppose I should just implement it as you describe, for now at least.  My hesitancy is due to the nagging idea that there's a better solution that hasn't quite come into focus for me yet, so I've been addressing other bits and pieces instead.  I think I'd like to think about and probably discuss it a little more, but the limited case of moving the needle (continuing the vinyl metaphor) when panning while paused should probably just be implemented asap.

PR #29, opened 2021-05-05
Settings panel

simonwiles:
So here's a "slide-over" settings panel as somewhere to put feature toggles and so on.  Some notes (in my usual prolix fashion):

* I don't know that I love the UI much at all, tbh.  I'm sort of imagining there may be more than two panels in this position (perhaps depending on the overall layout) which is why I stayed away from a single on/off button.  Will be interested in feedback.

* The layout/design of the the controls in the settings panel is tbd, of course.

* I think the playback controls should probably come out and be visible at all times.  I have some other ideas in this area, and will hopefully get to that soon.

* The theme changer can either a) stay in the final app; b) be dropped immediately; or c) stick around as a way of exposing alternatives for consideration, and then be dropped once a determination is made.  I'm not much bothered which, personally, I just wanted something on the panel for now, and I found it mildly entertaining as a way of experimenting with palettes myself :)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -37,6 +37,11 @@
   #keyboard-container {
     flex: 0 1 auto;
   }
+
+  #right-panel-wrapper {
+    position: relative;
+    margin-top: 1em;
+  }
 </style>
 
 <script>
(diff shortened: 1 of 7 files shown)


broadwell:
This looks great. The slideout panel should have some kind of border around it, but all in good time.

I also can imagine that we might have a few different settings panels that could slide out on this side (or even on the other side, though probably we won't go there): one for key mappings, one for emulation settings, one for advanced visualization displays (maybe), etc.

The theme switcher functionality is great to have. Even if it's ultimately not exposed to the user, we still might end up automatically switching color schemes to give visual cues about when the player is in interactive pianola mode vs. reproducing roll playback mode, with its considerably restricted interactivity.

PR #30, opened 2021-05-06
Details on active

simonwiles:
Here's the option to have the note names popup on active holes.  The PR is basically just some CSS and a new `$userSettings` store.  Not much to see here :)

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -59,21 +59,6 @@
         animation: mark-recede 0.5s ease-in-out;
         background-color: $hole-highlight-color;
         box-shadow: 0 0 5px $hole-highlight-color;
-
-        &[data-info]::after {
-          background-color: none;
-          color: white;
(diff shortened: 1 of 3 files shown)


broadwell:
Cool -- it's great to have the active note names feature available so we can get feedback on it.

Hmm, now I'm kind of wishing that the settings panel didn't obscure the play/pause control when open, just because it gets kind of cumbersome to change a setting like "Show details for active notes", close the panel, and then start playback. But this layout and functionality is all still pretty experimental, so it's probably best to wait until later to refine it.

simonwiles:
> Hmm, now I'm kind of wishing that the settings panel didn't obscure the play/pause control when open, just because it gets kind of cumbersome to change a setting like "Show details for active notes", close the panel, and then start playback. But this layout and functionality is all still pretty experimental, so it's probably best to wait until later to refine it.

Yeah, I mentioned that in the last PR.  Perhaps I'll just do it now.

PR #31, opened 2021-05-06
Separate playback controls

simonwiles:
Quick ad-hoc PR to pluck the playback controls out of the (newly-renamed) playback settings panel, and have them independent and always stuck to the bottom of the right-hand sidebar.  More to come here, but this can go in for now if it passes muster.


broadwell:
Yeah I think this layout is preferable, at least for now.

PR #32, opened 2021-05-07
Keyboard updates

simonwiles:
Here are some updates to the `<Keyboard/>` component.  Essentially it's a re-implementation of the CSS for the keys that makes it more-or-less fully scalable in response to the available width, and an 80% implementation of playability (including glissando).  I say 80% because it doesn't have (multi-)touch support yet.  I've just got the necessary ports opened on my laptop's NetDB registration so that I can host a dev. version that I can connect to from a touch device, so I'll clear this bit up fairly soon.  Oh -- there's also a change to the footer, although there's plenty still tbd there I suppose.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -53,6 +53,8 @@
   import {
     midiSamplePlayer,
     pianoReady,
+    startNote,
+    stopNote,
     stopAllNotes,
   } from "./components/SamplePlayer";
   import RollSelector from "./components/RollSelector.svelte";
@@ -204,7 +206,7 @@
     {/if}
   </div>
   <div id="keyboard-container">
-    <Keyboard keyCount="88" {activeNotes} />
+    <Keyboard keyCount="88" {activeNotes} {startNote} {stopNote} />
   </div>
   {#if !appReady}
     <div id="loading">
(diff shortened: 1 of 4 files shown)


broadwell:
The (more) responsive keyboard is great! I'm on the fence about the pop-up footer with transparent logos impinging on the keyboard, but it does save vertical real estate that way.

So I get one weird error when I stress the keyboard glissando feature by obnoxiously swiping left and right a bunch of times:
Screen Shot 2021-05-06 at 10 38 46 PM
It has happened on both Firefox and Opera. Oddly I can't get the Javatron to throw the same error, even though it's running pretty much the same sample playback code behind the keyboard. Not sure this is worth worrying about too much now, although you just know someone is going to find it while pretending to be Jerry Lee Lewis...

Good idea to test this with an actual multitouch device. The iPad emulators in Firefox and Opera can play single notes, although only as staccato -- the keys aren't held down for whatever reason. Firefox also throws some weird OSD error when I try to drag the roll in iPad mode, but let's not worry about that unless it happens with a real device.

broadwell:
Great, this passed all "swipe" tests after adding the `NaN` check.

I'm 100% fine with using this footer for the time being.

Thanks for reminding me that I can fire up the iPad emulator in Xcode. Oh looks like it needs to download a few gigs of updates before it can do that, surprise surprise.

PR #33, opened 2021-05-07
Vertical panning and wheeling

simonwiles:
This is the best I've been able to get this so far, then (looks awful simple when you look at the overall diff!).

* Vertical panning works nicely when the roll is stopped.  There could perhaps be a conversation to be had around the sensitivity settings.  The panning is using OpenSeadragon's native implementation, and it does expose some values we could easily adjust.

* The panning works very comfortably while playing on the Chromium-based browsers, and fails (but tolerably so) on FF.

* Ctrl+Mousewheel (on the roll image) scrolls the roll (I assume this maps to Cmd+Mousewheel on Macs -- it's only just occurred to me now to think about that).  The scroll increment is 10% of the viewport height per event, which feels comfortable to me.  We may have the whole scroll-direction inversion Macs vs. non-Macs thing here again.  Writing these notes now I also wonder whether overlay buttons could be provided to offer the same scroll up/down functionality without the mousewheel.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -198,7 +198,11 @@
     </FlexCollapsible>
     {#if appReady}
       <div id="roll">
-        <RollViewer imageUrl={currentRoll.image_url} {holesByTickInterval} />
+        <RollViewer
+          imageUrl={currentRoll.image_url}
+          {holesByTickInterval}
+          {skipToTick}
+        />
       </div>
       <FlexCollapsible id="right-sidebar" width="20vw" position="left">
         <TabbedPanel {playPauseApp} {stopApp} {skipToPercentage} />
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -344,7 +344,13 @@
   on:mouseleave={() => (showControls = false)}
   on:wheel|capture|preventDefault={(event) => {
     if (event.ctrlKey) {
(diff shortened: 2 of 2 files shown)


broadwell:
I really like how the scrolling interacts with the hole highlighting -- I think that makes this the best implementation so far.

Not sure how to improve the pan-during-zoom behavior; disabling it entirely seems undesirable. Maybe just increase the zoom speed/spring stiffness or whatever so that the zoom process completes faster, leading to less overlap with subsequent pans?

Panning while playing seems to work OK for Firefox on Mac, for what that's worth. On all browsers it can introduce a slight hiccup into the playback. Which we probably can live with -- it might serve as a subtle signal to the user that it's better to pan around while the roll is paused.

The CTRL+mousewheel panning direction on Macs seems intuitive to me (but then, I do use the "invert Y" setting a lot when gaming), and the 10% ratio seems to work well as a sort of "super-scroll" magnitude. Interestingly, the OS keymappings for accessibility-oriented zooming were set to CTRL+mousewheel, so I had to switch those around before it would work at all for me. Again, the long slippy pan/zoom times do introduce some annoyances, though -- if I release the CTRL key while it's still panning, some of the residual panning immediately converts to unwanted zooming (the usual mousewheel behavior). Not sure if we'd be able to finesse this away through event handling; increasing the speed of the pan/zoom in the OSD settings might suffice.

Anyway I think this is definitely a bit step in the right direction.

PR #34, opened 2021-05-11
Vertical panning buttons

simonwiles:
Some buttons for those who don't have a <kbd>ctrl</kbd>/<kbd>cmd</kbd> key (also better discoverability etc.).

A couple of things thrown up by this:

1) Previously the <kbd>ctrl</kbd>/<kbd>cmd</kbd> + mousewheel thing was moving in opposite directions depending on the direction of the roll.  This was not intentional on my part (!), but perhaps you noticed and liked it?

2) It's possible to scroll off the ends of the roll and into infinity, it seems (with buttons or mousewheel).

3) Not directly related to this PR, but I noticed while testing:  using the zoom functionality can cause the red line "tracker bar" indicator to get out of sync with where the actually playback position is.  This is a little unpolished.

I consider (2) and (3) to be fairly minor polishing issues that I'm inclined to wait until the PIs have seen all the recent changes before cleaning up.  Seem reasonable?

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -330,6 +330,18 @@
     openSeadragon.open(imageUrl);
   });
 
+  const panByIncrement = (down = true) => {
+    const viewportBounds = viewport.getBounds();
+    const imgBounds = viewport.viewportToImageRectangle(viewportBounds);
+    const delta = down ? imgBounds.height / 10 : -imgBounds.height / 10;
+    const centerY = imgBounds.y + imgBounds.height / 2;
+    skipToTick(
+      scrollDownwards
+        ? centerY + delta - firstHolePx
+        : firstHolePx - centerY + delta,
+    );
+  };
+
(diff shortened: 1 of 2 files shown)


broadwell:
This is great functionality, I think. I also feel like it's preferable for the <kbd>ctrl</kbd>/<kbd>cmd</kbd> scroll directions to be consistent across roll types.

I agree that the "infinite scroll" and inconsistent positioning of the red line with respect to the playback point can be finessed later.

A minor gripe is that the behavior of the scroll buttons vs. the zoom/fit buttons feels a bit inconsistent: maybe it should also be possible to repeatedly activate the zoom in/out buttons by holding them down, rather than having them fire on mouseup? Also, the magnitude of the scroll compared to the magnitude of zoom in/out seems like a bit of a mismatch to me -- but probably just because I already feel like the zoom in/out magnitude should be bigger, like 15-20% (unless holding down the buttons is possible, in which case it probably wouldn't matter as much).

PR #35, opened 2021-05-14
Store updates

simonwiles:
These are some updates to the Svelte stores that have been sitting in my working branches for a while and need to be merged to support some future changes.  In particular, the `media` store allows adapting layout and behaviour in Svelte components by using `{#if $media.narrow} <-- -> {/if}` blocks and so on.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -106,17 +106,17 @@
   const stopApp = () => {
     midiSamplePlayer.stop();
     stopAllNotes();
-    playbackProgress.set(0);
-    currentTick.set(0);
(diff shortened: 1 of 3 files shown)


simonwiles:
I've added an additional commit that creates a derived `sustain` store, as it seems to belong here rather than anywhere else.

broadwell:
Great! Cool to see that you've abstracted some of the store creation processes; otherwise it seems like that could get boilerplate-y pretty quickly. The store for media query data also seems like a good way to coordinate the responsive layout -- at least, I can't think of a better one.

PR #36, opened 2021-05-17
Skeuomorphic Pedals

simonwiles:
Here are the pedals, then.  I don't think there's too much to say about them.  I do think there might be some value in continuing to have the existing soft/sustain buttons, or some other less ambiguous indicators?  Feedback on aesthetics and behaviour very welcome :)


src/components/Keyboard.svelte
```diff
@@ -125,9 +127,28 @@
       }
     }
   }
+  svg.pedal {
```
broadwell, line 130: Based on extensive research (i.e., looking at photos of pianos on the internet), I think the following values will produce pedal positions that are a bit more true to life:
```
margin: 0 4%;
left: 38%
```

broadwell: Maybe also add a 1% bottom margin so that the shadows aren't truncated?
`margin: 0 4% 1%;`
This does somewhat increase the unused space at the bottom of the app, though.
Screen Shot 2021-05-17 at 2 57 34 PM
Screen Shot 2021-05-17 at 2 59 25 PM

broadwell: It turns out that the shadows are only truncated for Chrome/Opera on Mac OS, so it's probably better not to add a bottom margin and instead fix it some other way (or just leave it be).

simonwiles: Got it -- changes applied.

broadwell:
﻿This is great. The pedal SVG appearance maybe can be finessed a bit more, but I think it's already >90% perfected. A couple of other outstanding refinements to consider eventually:
﻿- Pedals could scale more appropriately relative to the keyboard size at extreme window dimensions.
﻿- The bottom/shadow of the non-depressed pedals are truncated in MacOS Chrome/Opera -- but not on Firefox or Safari, and the depressed pedals look fine in all browsers.

simonwiles:
> ﻿- Pedals could scale more appropriately relative to the keyboard size at extreme window dimensions.
Addressed by a36baf1 in #37.

broadwell:
Just adding here that I agree it's probably preferable to keep the Soft and Sustain buttons in addition to the pedals. Having the buttons act as "lock" (toggle) buttons while the pedals are only active while being held down (which is also how the mapped QWERTY keys will act, presumably) seems like a sensible way to differentiate their functionality.

PR #37, opened 2021-05-18
CSS Refactoring and Fixes

simonwiles:
This is a little clutch of CSS-related commits.

1) There's a bit of a reorganization: base styles are moved out to `base.scss` and `styles.scss` is just for `@import` statements; the overlay button styles are abstracted out of the `<RollViewerControls/>` component so as to be reusable.  I'd like to unify the styles with the buttons for the sidebars, but haven't done so yet.

2) I've introduced a system for keeping track of `z-index` values, based on [this article](https://www.smashingmagazine.com/2014/06/sassy-z-index-management-for-complex-layouts/).  It may be (mild) overkill here, but I've used this system before and found it worth the (fairly minimal) effort.

3) There's what looks like a bug to me in Snowpack whereby it sometimes attempts to import SCSS partials as Javascript modules (and duly fails).  Using relative paths for the SCSS imports guards against this issue.

4) There's a small fix for the hole overlay hover effect on FF -- on active notes the "flag" (as I heard you call it today :)) gets offset because of the note name above the hole (even if the latter is hidden).  a36baf1 fixes this.

5) The last commit fixes the issue with the pedals getting out of proportion on extra large or small viewport widths.  I was too hasty dropping this `width` value -- I should simply have adjusted it like this.


broadwell:
This is nice... The list-based system for z-index values makes a lot of sense. And the pedals do respond much better to container size changes now.

PR #38, opened 2021-05-19
Overlay Keyboard

simonwiles:
I don't think there's too much to be said about this -- hopefully it "just works".

* I had been intending a second button to slide the keyboard out but *not* overlay it on the roll image (i.e. just to turn the keyboard off) -- this may still be appropriate.

* Also the button styling isn't ideal.  There's a bit of fragmentation between the different overlay buttons that could (and probably should/will) be addressed at some point; in this case I didn't want to use the CSS-only approach that I used for the collapsible sidebars because I really wanted the keyboard component destroyed and out of the DOM altogether when not on screen.

* The icon is not ideal either, but probably acceptable?  I toyed with creating something new and entirely bespoke, but decided to move on to other things instead, for now at least.

* There are also minor side issues like the white outline that has rounded corners on FF but not on other browsers.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -38,6 +38,14 @@
     flex: 0 1 auto;
     user-select: none;
   }
+
+  #keyboard-overlay {
+    position: absolute;
+    z-index: 23;
+    bottom: 0;
+    width: 100%;
+    opacity: 0.5;
+  }
 </style>
 
 <script>
@@ -72,6 +80,7 @@
   let currentRoll;
   let previousRoll;
   let holesByTickInterval = new IntervalTree();
+  let overlayKeyboard = false;
 
   const buildHolesIntervalTree = () => {
     const { ROLL_TYPE, FIRST_HOLE, IMAGE_LENGTH, holeData } = $rollMetadata;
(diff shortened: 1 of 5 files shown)


broadwell:
I'm a big fan of this feature. Much of my enthusiasm is probably just because it works at all (and so smoothly) and looks cool, but I think the extra layout affordances it provides likely will prove beneficial, particularly as we add more content to the sidebars.

I agree that eventually it will be nice to have the option of hiding/dropping the keyboard entirely.

I'm fine with the current suite of buttons for now. I suspect that we may end up doing a general refresh of the buttons' appearance and behavior eventually to provide better touch device support. At present, you kind of need to know where several of the buttons are to use them (at least on the iPad simulator). Having them pop out when the relevant region or component is tapped would be fine, I think, but at present that behavior doesn't seem very consistent.

simonwiles:
>I'm fine with the current suite of buttons for now. I suspect that we may end up doing a general refresh of the buttons' appearance and behavior eventually to provide better touch device support. At present, you kind of need to know where several of the buttons are hiding to use them (at least on the iPad simulator). Having them pop out when the relevant region or component is tapped would be fine, I think, but at present that behavior doesn't seem very consistent.

Agreed.  Proper touch support is on my list for most of the components, tbh, and that will also be an opportunity to unify the two different approaches that currently exist side-by-side in the code base.  My default is always to use the minimum JS possible as a first option, but then some things do need to be done from within JS (or Svelte in this case) and now I've got two different solutions to maintain independently...  TL;DR I expect to be rewriting the collapsible sidebar toggles in Svelte and using a single extendable component with 100% consistent styling and behaviour (including touch) for all the various use cases.

PR #39, opened 2021-05-19
Tempo Map

simonwiles:
Tempo Map PR as discussed.  Storing the (user-controlled) tempo value as a coefficient seems desirable, but that can of course be separate from the way it is displayed to users (i.e. we could still convert back to bpm or something if such a conversion seems logically consistent and valid).

See also 7fa3e72 below.  Are we happy just to accept that `midi-player-js` only returns integers for tempo events and move along?  Or should this be on our issue list?

### DIFF
--- a/src/components/SamplePlayer.js
+++ b/src/components/SamplePlayer.js
@@ -36,6 +36,7 @@ const getTempoAtTick = (tick) => {
   while (tempoMap[i][0] <= tick) {
     [, tempo] = tempoMap[i];
     i += 1;
+    if (i >= tempoMap.length) break;
   }
   return tempo;
 };
--- a/src/components/SamplePlayer.js
+++ b/src/components/SamplePlayer.js
@@ -14,16 +14,6 @@ import {
 
 const midiSamplePlayer = new MidiPlayer.Player();
 
-const updatePlayer = (fn) => {
-  if (midiSamplePlayer.isPlaying()) {
-    midiSamplePlayer.pause();
-    fn();
-    midiSamplePlayer.play();
-    return;
-  }
-  fn();
-};
-
 let softPedalOn;
 let accentOn;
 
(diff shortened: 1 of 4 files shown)


src/components/SamplePlayer.js
```diff
@@ -36,12 +26,36 @@ volume.subscribe(({ master, right, left }) => {
...
+  let i = 0;
+  while (tempoMap[i][0] <= tick) {
+    [, tempo] = tempoMap[i];
+    i += 1;
+    if (i >= tempoMap.length) break;
+  }
+  return tempo;
+};
+
+const updatePlayer = (fn = () => {}) => {
```
broadwell, line 44: As I recall, applying a tempo change while the player is running works fine and doesn't cause any known complications, so wouldn't this be slightly more efficient (if only slightly)?
```
const updatePlayer = (fn = () => {}) => {
  if (midiSamplePlayer.isPlaying()) {
    midiSamplePlayer.pause();
    fn();
    midiSamplePlayer.play();
  } else {
    fn();
  }
  midiSamplePlayer.setTempo(getTempoAtTick(get(currentTick)) * tempoRatio);
};
```

simonwiles: I think I did this because the prototype also seems to pause the player before changing the tempo?  (Don't remember testing it without, tbh.)

It might be more efficient in the case where no function was passed (meaning there was no need to stop the player at all), perhaps, in which case that would need to be checked in the initial condition, of course.

Otherwise it's the same except that in the original implementation the player is paused for the extra (JS engine) tick while the tempo is set (the only time this occurs at present is when skipping, so it doesn't create any additional stutter), whereas with this change the player would be started and *then* the tempo would be updated.  The former still seems preferable to me.

If stopping the player to update the tempo really isn't necessary, and you're concerned therefore about the invocation at line 55, a better option might be to simple call `midiSample.setTempo()` directly there instead.

Or perhaps I'm not understanding properly?

broadwell: Never mind, pausing on tempo change actually was the preferred behavior with the prototype, and I don't see any reason to change it here!

broadwell:
Works well for me!

PR #40, opened 2021-05-20
Keyboard Shortcuts

simonwiles:
Here are the basic keyboard shortcuts we need.  A couple of notes:

* increment and decrement values can probably be tweaked, and perhaps made configurable?
* the "hints" to indicate available keys leave something to be desired.  I spent some time trying to come up with something but ultimately failed and decided to move on.
* One thing I experimented with re. the above was increment and decrement buttons -- I feel like a reset button is probably appropriate (on a control-by-control basis and/or for all settings at once), and perhaps this is worth considering in the same context.
* This PR (and the whole concept up to this point) has been focused on key controls to interact with playback, but of course we could start adding additional shortcuts to this framework very easily -- perhaps the spacebar to play/pause, keyboard panning and zooming controls, etc.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -54,7 +54,8 @@
   import IntervalTree from "node-interval-tree";
   import {
     pedalling,
-    volume,
+    bassVolume,
+    trebleVolume,
     tempoControl,
     playbackProgress,
     activeNotes,
@@ -140,7 +141,8 @@
     appReady = false;
     stopApp();
     tempoControl.reset();
-    volume.update((val) => ({ ...val, left: 1, right: 1 }));
+    bassVolume.reset();
+    trebleVolume.reset();
     holesByTickInterval = new IntervalTree();
   };
 
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -74,6 +74,7 @@
   import RollDetails from "./components/RollDetails.svelte";
   import RollViewer from "./components/RollViewer.svelte";
   import Keyboard from "./components/Keyboard.svelte";
+  import KeyboardShortcuts from "./components/KeyboardShortcuts.svelte";
   import TabbedPanel from "./components/TabbedPanel.svelte";
   import Notification, { notify } from "./ui-components/Notification.svelte";
   import FlexCollapsible from "./ui-components/FlexCollapsible.svelte";
@@ -242,4 +243,5 @@
     </div>
   {/if}
 </div>
+<KeyboardShortcuts />
 <Notification />
--- a/src/components/KeyboardShortcuts.svelte
+++ b/src/components/KeyboardShortcuts.svelte
@@ -21,11 +21,11 @@
         break;
 
       case keyMap.VOLUME_UP:
-        $volume.master += 0.1;
+        $volume.master = Math.round(($volume.master + 0.1) * 10) / 10;
         break;
 
       case keyMap.VOLUME_DOWN:
-        $volume.master -= 0.1;
+        $volume.master = Math.round(($volume.master - 0.1) * 10) / 10;
         break;
 
       // no default
--- a/src/components/KeyboardShortcuts.svelte
+++ b/src/components/KeyboardShortcuts.svelte
@@ -1,4 +1,5 @@
 <script>
+  import { get } from "svelte/store";
   import { pedalling, volume, tempoControl } from "../stores";
 
   const keyMap = Object.freeze({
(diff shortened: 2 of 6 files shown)


broadwell:
Great! Some reples:
- As for making the increment/decrement step sizes configurable, I think we should approach this in the same way as the button/key remapping feature, i.e., only pursue it if directly requested to do so.
- We can continue to refine the appearance of the keyboard mapping hints on the buttons. We also should eventually explain the availability of the <kbd>ctrl</kbd> and <kbd>shift</kbd> modifiers somewhere, but I'd prefer to do this on a separate help page, rather than having it contribute to clutter on the main page. But the PIs may have other opinions.
- I agree that a "Reset All" button will be a good addition to the UI. Per-control reset and Increment/decrement buttons might be a bit too fiddly, especially with the sliders right there, but we'll see.
- Space bar for play/pause would be welcome, I think. Not sure about the other mappings -- maybe wait and see if they're requested?

FYI it turns out that the reduced volume resolution/responsiveness vs. the prototype is due entirely to the use of 2 sample velocities rather than 4 (or more). The sample player doesn't seem to be able to interpolate meaningfully between, say, .2% and .6% with such a coarse level of base resolution. We'll probably need to provide at least the option of using more samples. Also, to reduce the distortion heard at volume levels above 1, it really helps to reduce the gain on all of the sample volumes, as done in the prototype:
```
piano.strings.value = -15;
piano.harmonics.value = -10;
piano.pedal.value = -10;
piano.keybed.value = -10;
```

simonwiles:
Thanks, this is all very interesting and useful.

> We can continue to refine the appearance of the keyboard mapping hints on the buttons. We also should eventually explain the availability of the ctrl and shift modifiers somewhere, but I'd prefer to do this on a separate help page, rather than having it contribute to clutter on the main page. But the PIs may have other opinions.

I've been imagining pop-up help on a control-by-control basis (probably triggered by a little 🛈 symbol or similar) -- such help popups could potentially also be stepped through in an `intro.js` -style tour, too.

> I agree that a "Reset All" button will be a good addition to the UI. Per-control reset and Increment/decrement buttons might be a bit too fiddly, especially with the sliders right there, but we'll see.

I reckon a small ⟲ button at the end of each slider would probably be all right, personally.  Might try it out and present it for consideration.

> FYI it turns out that the reduced volume resolution/responsiveness vs. the prototype is due entirely to the use of 2 sample velocities rather than 4 (or more). The sample player doesn't seem to be able to interpolate meaningfully between, say, .2% and .6% with such a coarse level of base resolution. We'll probably need to provide at least the option of using more samples. Also, to reduce the distortion heard at volume levels above 1, it really helps to reduce the gain on all of the sample volumes, as done in the prototype:

Ah, okay, that does figure.  What do you reckon, then?  Implement the reduced gain, default to four sample velocities?  It does bump up the memory usage of the app by about 50%, but we're already in fairly high-end device territory (relative to general standards, perhaps not Stanford / Bay Area standards).

A second look at the way velocities are computed in general may be warranted.  With a volume multiplier of `1`, we're not making much use of the dynamic range available anyway.  I `console.log`ed the `modifiedVelocity` value every time a note is played, and depending on the roll it varies between about `0.3` and `0.5`.  Perhaps there's something we can do there?  The master volume is of questionable utility anyway, imo -- device-level volume controls are available for overall output.  Something to consider, perhaps.

simonwiles:
Oh, and regarding the increments/decrements:  leaving aside the issue of making them user-configurable, do you think the defaults make sense (and the default ranges, too, for that matter)?  Perhaps we could tune them a little ourselves (and thus forestall any requests for making them configurable, too!).

broadwell:
> I've been imagining pop-up help on a control-by-control basis (probably triggered by a little 🛈 symbol or similar) -- such help popups could potentially also be stepped through in an intro.js -style tour, too.

This seems like a great idea to me. Maybe we could even have a popup near the top of the controls that lists all of the key mappings, so it's not necessary to click/mouseover each control to see its mapping.

> I reckon a small ⟲ button at the end of each slider would probably be all right, personally. Might try it out and present it for consideration.

Ok cool, if it's something relatively small and unobtrusive then that seems fine to me.

> Oh, and regarding the increments/decrements: leaving aside the issue of making them user-configurable, do you think the defaults make sense (and the default ranges, too, for that matter)? Perhaps we could tune them a little ourselves (and thus forestall any requests for making them configurable, too!).

I think the increments for both tempo and volume are already pretty well tuned, especially with the <kbd>ctrl</kbd> and <kbd>shift</kbd> modifiers. The max for both the tempo and volume ranges probably could be reduced, though, maybe even just to 2x for both.

Re: samples, velocities and volume -- I do think the 4 sample velocities and the reduced gain amounts for the samples are acceptable defaults. Another question is whether we want to provide the option of changing the number of samples, which requires recycling the entire piano component upon modification (which is still fairly straightforward; the prototype does it, albeit not all that elegantly).

I also agree that we should do a more comprehensive review all of the parameters and mechanisms for determining playback loudness. The note velocity levels for the expressionized rolls do tend to be limited to a pretty narrow range (well, at least for the "official" expression MIDI files; the ones I'm generating now with the most recent midi2exp code are worryingly out of whack), but this is presumably intentional/by design, so we'll want to maintain the ability to hear them as such. But we also could expose more of the gain settings in the app, which would provide something like an in-app proxy for device-level volume controls.  
And I think eventually we'll find ourselves wanting to implement most of the velocity-massaging features that Peter P demonstrated: dynamic range adjustments (which could include maximizing the range of velocity levels), and the "offset" and "generator" modifications to make the output sound as "pianesque" as possible.

PR #41, opened 2021-05-21
Store and Player Refactoring

simonwiles:
Stores are renamed and reorganized, and the sample player is now a `<SamplePlayer/>` svelte component.  There should be no behaviour changes here (except for some small performance improvements), but the code is neater and tidier, and further gains from this will be realized when the expression modification code is in.


broadwell:
These all look like productive changes to me. It's great that the sample player has been fully Svelte-ified now. The syntax modifications around updating/reacting to the store changes all seem to bend towards greater clarity, or at least less boilerplate.
I suppose I could quibble over whether `OnOff` is preferable to something else, like `Toggle`, and whether `Coefficient` could be replaced by something like `Mult`in the variable names, but I don't feel strongly enough about either to hold up the PR.

PR #42, opened 2021-05-21
Expression Modifications

simonwiles:
Implements the expression modification toggles.

* the settings panels are renamed to `<BasicSettings/>` and `<AdvancedSettings/>`

* "Play Expressions", "Use Roll Pedaling", and "Emulate Roll Acceleration" are implemented.

* "Play Expressions" and "Use Roll Pedaling" are disabled when playing an 88-note roll.

* The state of "Use Roll Pedaling" and "Emulate Roll Acceleration" are monitored and relevant app state is automatically updated when they change.
  * Activating appropriate pedaling when "Use Roll Pedaling" is enabled in the middle of a roll is incomplete pending the implementation of a map of pedal events.


src/components/SamplePlayer.svelte
```diff
@@ -25,6 +28,7 @@
   });
 
   const DEFAULT_NOTE_VELOCITY = 33.0;
```
broadwell, line 30: Probably should be 64 (can be fixed later).

broadwell:
This is great. I like the styling of the settings panel, too.

Also, I just remembered one reason why the naming of the `useMidiTempoEvents...` variable deviates from the text of the settings option ("Emulate Roll Acceleration") in the Javatron prototype: it actually has code to compute an alternative tempoMap for the roll using a more realistic time-based (rather than roll distance-based) method. So there actually could be two different settings: one to toggle the MIDI tempo events, and one to toggle the alternative tempoMap.  
In practice, though, I think we should just compute our own tempo map for each roll at load time because the methods used for both the Welte and 88-note rolls are (subtly) problematic, in which case we'd always disregard the tempo events in the MIDI file, and the "Emulate Roll Acceleration" setting then would toggle between applying no acceleration at all (tempo always = 60, or whatever the default is), and using the tempoMap that has been computed in-app.

src/components/SamplePlayer.svelte
simonwiles: Yeah, it's actually 50, of course, once the velocities are divided by 100 and not 127, and the fix is incoming in the next PR :)

simonwiles:
Hmm, yes, I forgot there's more going on here.

> In practice, though, I think we should just compute our own tempo map for each roll at load time

As opposed to baking it into the MIDI files?  How does the problematic tempo adjustment play out in the mp3s?

broadwell:
The problematic acceleration is present in the MP3s, although it must be said that the difference from the "corrected" tempo map is basically imperceptible for the Welte red rolls (unless they're extremely long). It is a bit more noticeable in MP3s for the Garcia-Sampedro rolls because the implementation for those rolls actually manages to use the wrong constant. On the other hand, just listening to the playback of a non-performed pianola roll is always a somewhat unsatisfactory experience, hence the Pianolatron!

We certainly could generate corrected MIDI files as well. Ideally we'd just fix the C++, but I'm fairly certain that the idiosyncrasy of the roll acceleration emulation code contributed to the snafu with the Garcia-Sampedro acceleration -- so that should give some idea of how gnarly it is.

simonwiles:
> We certainly could generate corrected MIDI files as well. Ideally we'd just fix the C++, but I'm fairly certain that the idiosyncrasy of the roll acceleration emulation code contributed to the snafu with the Garcia-Sampedro acceleration -- so that should give some idea of how gnarly it is.

Okay, got it ;)

PR #44, opened 2021-05-23
Misc Fixes

simonwiles:
This wound up a bit more unruly (a bit less ruly?) than I'd intended, but here's what's here (in order):

* URL fix for the little CIDR "i"
* A couple of linting commits
* Fix for velocity computations (`midi-player-js` returns a value between 0 and 100, not between 0 and 127)
* All the rest are ultimately targeted at preventing panning/scrolling off the ends of the roll, and handling associated playback behaviour (especially when beyond the range of the actual note holes).  This ended up involving a minor refactor of some of the `<SamplePlayer/>` internals to expose a more optimal API to the main app component.  The other user-facing change that got caught up in this process is that the roll no longer automatically returns to the beginning when the playback reaches the end.  This was already something I was planning on changing, as it seems a much better UX to me this way.  Dissenting opinions welcome, though!

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -211,7 +211,6 @@
       stopNote,
       stopAllNotes,
     } = samplePlayer);
-    midiSamplePlayer.on("endOfFile", stopApp);
   });
 
   $: if (currentRoll !== previousRoll) loadRoll(currentRoll);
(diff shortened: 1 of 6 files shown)


broadwell:
I think these are all very worthwhile refinements/fixes that streamline the codebase as well as the app's functionality.

I agree that stopping the roll and playback at the end of the playing section is the right UI. I think that some (most?) player pianos had an auto-rewind mechanism, but see no compelling reason to replicate this behavior.

PR #45, opened 2021-05-25
Increase available velocities

simonwiles:
This PR adds 120 additional `.mp3` sample files to the repo, defaults the `@tonejs/piano` to `velocities: 4`, and adjusts the volume levels for the individual components.


simonwiles:
As mentioned I'd advocate for bumping the default `$volumeCoefficient` significantly (x2?) along with this.

broadwell:
I'd be in favor of defaulting the `$volumeCoefficient` to something like `1.5`. `2` makes the upper registers sound pretty shrill to my ears. Also fwiw, one motivating factor behind my preference for the 4 velocity levels is that it makes the app's playback sound to me more like the MP3 performances of the rolls produced in Ivory II, which on the whole seem to go for a richer but slightly muted sound quality.

Exposing the available velocity sample level settings, as well as the gain settings for the strings and maybe the other samples as advanced options seems like a worthwhile stretch goal, along with perhaps implementing some of the range normalization and register or even note-based adjustment features from Peter P's app.

PR #46, opened 2021-05-26
Fixes

simonwiles:
* Remove superfluous calls to `piano.pedalDown()` and `piano.pedalUp()` in `<SamplePlayer/>` component.
* **Wrap SVG pedals in `<div/>`s, and animate the `<div/>`s for performance on Mac OS blink-based browsers**
* Change default volume level to 1.5
* Release`<Keyboard/>` keys when panning away

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -125,6 +125,7 @@
   const skipToTick = (tick) => {
     if (tick < 0) pausePlayback();
     $currentTick = tick;
+    activeNotes.reset();
     updatePlayer(() => midiSamplePlayer.skipToTick($currentTick));
   };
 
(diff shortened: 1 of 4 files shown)


simonwiles:
Haven't set the volume back to 1 for the 88-note rolls, though -- would you like to see that added, or leave it as it is for now?

broadwell:
I'm OK with leaving the volume at 1.5 for the 88-note rolls for now -- we might want to confer with the PIs about this first, anyway.

The pedaling and key highlighting changes look good. Unfortunately, wrapping the pedal SVGs in `<div>` tags doesn't seem to have much of an effect on the MacOS+Chrome playback problems. 😞

PR #49, opened 2021-06-07
CSS updates

simonwiles:
These have been sitting around for a couple of weeks waiting to go in.  This PR includes:

* The gradient background
* A small (probably pointless) adjustment to the range slider handles
* Styling for the keyboard shortcut indicators

My only concern is with the latter -- it does sort of make it look like they should be clickable, and they're not.  Dunno what you think.  I don't think I reckon it's a show-stopper?  (One possible solution would be to make it so they *are* clickable, of course, but I'm not sure it's really worth it?)

### DIFF
--- a/src/components/AdvancedSettings.svelte
+++ b/src/components/AdvancedSettings.svelte
@@ -1,6 +1,6 @@
 <style lang="scss">
   div#settings-panel {
-    background: var(--background-color);
+    @include background;
     height: 100%;
     left: 0;
     position: absolute;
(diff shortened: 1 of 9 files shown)


broadwell:
I like the new style for the range sliders.

Regarding the keyboard shortcut indicators: there's certainly a case to be made that they *should* be clickable; some users may appreciate being able to fine-tune the tempo and volume levels that way instead of fiddling with the slider. If we take that route, though, it may be desirable to use styling that makes it a bit more obvious that they're increment/decrement buttons in addition to keyboard mapping indicators (hopefully without obscuring either valence).

In the meantime, it's worth getting this in front of the PIs.

PR #50, opened 2021-06-09
Hole highlighting for both roll types; get MIDI numbers from JSON files

broadwell:
Building the holes interval tree and generating overlays for 88-note rolls is actually fairly straightforward: the image processing code already inverts the hole pixel coordinates on the Y axis for bottom-up scrolling rolls, which simplifies the arithmetic a bit.

Additionally, the backend data processing pipeline has advanced to the point that hole data, including note MIDI numbers, can be generated correctly for all of the test rolls, including the 88-note rolls, so now we can use the MIDI numbers provided in the hole reports (and subsequently incorporated into the roll JSON data) when assigning note names.


simonwiles (approved):

simonwiles:
It's great to see the hole highlighting on the 88-note rolls now.  I've added a couple of commits to satisfy `prettier` and `eslint` (that's on me, sorry -- I should set up CI or at least some kind of `husky` / pre-commit thing to enforce and/or automatically take care of that stuff).  I also have some questions about the JSON, but they don't need to hold this up.  I *will* wait to merge until after we've spoken this afternoon, though.

PR #51, opened 2021-06-09
Button for hiding/showing the Keyboard component

simonwiles:
These commits have been sitting un-PR'd for a couple of weeks too, and should probably have been part of the last PR, sorry.  Do you find this UI preferable?


broadwell:
This works fine for me, and the buttons look great!

PR #52, opened 2021-06-18
Add playable files and metadata for newly processed rolls

broadwell:
The sheer number of title/label entries sort of overwhelms the drop-down list, but the actual format and quality of the metadata seems workable.

Most of the handful of new rolls I've spot-checked actually seem to play as well as expected. I'm already compiling a list of problematic rolls, which perhaps also should be shared/crowdsourced with the PIs.

Further refinement of the roll image processing logic needs to be done regardless, but I think it's worth getting this greatly expanded set of rolls in front of the PIs now -- which also will free the devs from having to listen to the original 7 test rolls on perpetual repeat.


simonwiles (changes requested):
There seem to be two MIDI files missing?  (`pk349zj4179` and `xy736dn5214`) -- the JSON is present, and they're in `catalog.json`.  Are they available, or should we remove the rolls from the catalog and `json/` folder?

```
% for i in json/*; do stat midi/${${i##json/}%%.json}.mid > /dev/null; done;
stat: cannot statx 'midi/pk349zj4179.mid': No such file or directory
stat: cannot statx 'midi/xy736dn5214.mid': No such file or directory 

% for i in $(jq -r ".[] | .druid" catalog.json); do stat midi/$i.mid > /dev/null; done;
stat: cannot statx 'midi/pk349zj4179.mid': No such file or directory
stat: cannot statx 'midi/xy736dn5214.mid': No such file or directory 
```

broadwell:
Thanks for catching the missing MIDI files.
`pk349zj4179` has been added in the latest commits -- it has a malformed roll type string in its MODS record (ugh) that was keeping it from being processed, but it's a legit Welte red roll.
`xy736dn5214` is actually a 65-note pianola roll from the García-Sampedro collection -- the only one of this type, it seems. There actually is code in the roll parsing repo to handle 65-note rolls, but it hasn't been integrated into the new workflow yet, so for not it's probably best to exclude this roll. The JSON file and `catalog.json` entry for it have been removed.

simonwiles (approved):

simonwiles:
Awesome, thanks.  It's so nice to have so many more rolls available to play :)  I'll try to get at least an filtered select widget together asap.

PR #53, opened 2021-06-22
<FilteredSelect/> component for <RollSelector/>

simonwiles:
There are a couple of "Select2"-type components out there for Svelte that can be `npm install`ed (or `yarn add`ed, I suppose), but they're all a bit heavy and over-featured for this case, none had a really perfect UX, and most importantly all the ones I could find were lacking in at least one important respect, so would have needed some kind of hacking anyway.  This is written from scratch to meet our needs specifically.

It should be fully keyboard accessible/driveable, and will accept multiple "words" (e.g. type "bee lam").  I think it's pretty robust at this point (with the caveats listed below).  Please let me know if there are any UI/UX problems or niggles.

A couple of improvements/additions remain outstanding:
* Roll-type filtering (I think I know how I'd like to do it)
* Stripping accents for searching (as usual with these kinds of tasks -- and I've done it a tonne of times in a tonne of different programming environments, as you might imagine! -- the challenging part is the marking-up of the matches in context)
* I've not done any Safari or iPad / touchscreen testing at this point.

The first commit represents a lot of dev. work but my commits got really messy and I didn't have the energy to tidy them so I just squashed the mess into one commit last night and the rest of the commit are iterations from there.

### DIFF
--- a/src/components/RollSelector.svelte
+++ b/src/components/RollSelector.svelte
@@ -1,19 +1,13 @@
-<style>
-  select {
-    padding: 0.25em 0;
-    width: 100%;
-  }
-</style>
-
 <script>
-  // bundling this, for now at least
+  import FilteredSelect from "../ui-components/FilteredSelect.svelte";
   import catalog from "../assets/catalog.json";
 
   export let currentRoll = catalog[Math.floor(Math.random() * catalog.length)];
 </script>
 
-<select bind:value={currentRoll}>
-  {#each catalog as roll}
-    <option value={roll}>{roll.title}</option>
-  {/each}
-</select>
+<FilteredSelect
+  items={catalog}
+  bind:selectedItem={currentRoll}
+  labelFieldName="title"
+  searchFieldName="title"
+/>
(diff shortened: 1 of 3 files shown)


broadwell:
This is a very cool component. It worked fine in all of the Mac browsers I tried, including Xcode's iPad emulator.

Incidentally, the transient bug whereby the new roll image fails to load (but the metadata and MIDI playback changes) after selecting a new roll persists after this change. It seemed unlikely to be tied directly to the roll selector, and this confirms that. I'll also note anecdotally that it seems to happen more frequently if not exclusively in Firefox (on Mac), rather than Chrom*-based browsers.

simonwiles:
Thanks!

> Incidentally, the transient bug whereby the new roll image fails to load (but the metadata and MIDI playback changes) after selecting a new roll persists after this change. It seemed unlikely to be tied directly to the roll selector, and this confirms that. I'll also note anecdotally that it seems to happen more frequently if not exclusively in Firefox (on Mac), rather than Chrom*-based browsers.

I've not worked out what's causing this yet (not looked very hard yet), so if you do notice anything that seems relevant please make a note and let me know.  I see it (or something phenotypically similar, at least) occasionally when working on a local dev. server, but I've not seen it happen on a production build.

PR #54, opened 2021-06-22
Add roll label (ex: 4668 Princesa) to catalog.json, correct errors for 3 rolls

broadwell:
These updates include all corrections to known playback/roll viewing problems noticed in the expanded roll set that can be accomplished via changes to their JSON and MIDI files (so far). Also, the addition of the roll label data to catalog.json allows it to be used to enhance the roll search/selector to incorporate this data in some way, as requested by the PIs.


simonwiles:
Rebased and merged -- how was the change to `vj052cw2158` effected?  Was it a manual intervention, or is it a function of something you were able to address in the roll-parser code?

broadwell:
Thanks! For `vj052cw2158`, I just Gimped the offending thread out of the tiff, which of course introduces a non-tracked derivative into the workflow, posing similar challenges to the reproducibility of the processing workflow that a roll editing tool would. But I figured that throwing it in this time wouldn't be harmful.

Modifying the image parsing code would be a more sustainable fix given the current workflow (and with no way of easily replacing/versioning the images in the SDR), but the particular placement of this spurious hole -- right at the point at which the roll reaches its full width -- makes it oddly tricky to catch, especially without running the risk that the change would cause the parser to invalidate legitimate holes in other rolls.

simonwiles:
> For vj052cw2158, I just Gimped the offending thread out of the tiff, which of course introduces a non-tracked derivative into the workflow, posing similar challenges to the reproducibility of the processing workflow that a roll editing tool would. But I figured that throwing it in this time wouldn't be harmful.

Got it -- just wanted to make sure I knew the state of play.  I was wondering whether checking the distance between the first two holes might be a suitable heuristic?  It seems that the most common case is one stray/random hole and then a long wait for the piece to begin.  It might be possible to catch 9/10 of the problems by simply dropping the first hole if there's a gap of greater than `{x}` ticks to the second hole?

broadwell:
Yeah, as you note, your suggested fix would catch the vast majority of these problems -- and presumably run very little risk of invalidating legitimate music/control holes.
In this particular case, though, it's odd to me that a hole so far to the right edge of the roll was identified as a music/control hole at all -- based on my (obviously incomplete) understanding of the parsing process, that location should fall beyond the edge of the virtual tracker bar and thus not be classified as a music/control hole, which would mean that it would not be used as the `FIRST_HOLE` location for the hole report and MIDI playback due to the changes in [this PR](https://github.com/sul-cidr/roll-image-parser/pull/8) -- which were merged into the code I used when generating the JSON and MIDI files.
But obviously the parser code is not doing what I think it should for this roll, and in any case, I've noticed other rolls on which the spurious holes were only vertical outliers, i.e., they fell within the range of the virtual tracker bar, so your suggested fix would be needed to handle them.
This all could be something for a C++-savvy (or C++-enthusiastic) student developer to look into, much like the hole grouping issue. Or we could just do it ourselves.

PR #55, opened 2021-06-23
Highlight all holes

simonwiles:
Quick little PR to add an option to highlight all the holes at once, following our discussion yesterday.  I continue to doubt the usefulness (or even the relevance) of this, but it's simple and cheap enough to implement, so here it is.

### DIFF
--- a/src/components/AdvancedSettings.svelte
+++ b/src/components/AdvancedSettings.svelte
@@ -81,8 +81,12 @@
   <fieldset>
     <legend>Visualization Settings</legend>
     <div>
-      Show details for Active Notes:
+      Show Details for Active Notes:
       <input type="checkbox" bind:checked={$userSettings.activeNoteDetails} />
+    </div>
+    <div>
+      Highlight All Holes:
+      <input type="checkbox" bind:checked={$userSettings.showAllHoles} />
     </div>
   </fieldset>
 
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -252,6 +252,8 @@
       );
       rect.setAttribute("width", width + padding * 2);
       rect.setAttribute("height", height + padding * 2);
+      rect.setAttribute("rx", 10);
+      rect.setAttribute("ry", 10);
       rect.addEventListener("mouseover", () => {
         if (marks.map(([_hole]) => _hole).includes(hole)) return;
         viewport.viewer.removeOverlay(hoveredMark);
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -241,14 +241,17 @@
         "rect",
       );
       const { x: offsetX, y: offsetY, w: width, h: height } = hole;
+      const padding = 10;
 
-      rect.setAttribute("x", offsetX);
+      rect.setAttribute("x", offsetX - padding);
       rect.setAttribute(
         "y",
-        scrollDownwards ? offsetY : imageLength - offsetY - height,
+        scrollDownwards
+          ? offsetY - padding
+          : imageLength - offsetY - height - padding,
       );
-      rect.setAttribute("width", width);
-      rect.setAttribute("height", height);
+      rect.setAttribute("width", width + padding * 2);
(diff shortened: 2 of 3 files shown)


src/components/RollViewer.svelte
```diff
@@ -241,14 +241,17 @@
         "rect",
       );
       const { x: offsetX, y: offsetY, w: width, h: height } = hole;
+      const padding = 10;
```
broadwell, line 252: This certainly increases the dimensions of the `rect` elements, which is useful for highlighting, but wrt the commit message, on my browsers at least the change doesn't seem to have increased the range of the hover tripwire around the holes, if that is indeed the intention. I still need to place the mouse pointer within a quite narrow rectangle that is seemly defined by the actual hole locations in order for the highlight rectangle + info flag popup to appear.

src/components/RollViewer.svelte
```diff
@@ -132,6 +132,14 @@
         transform: none;
       }
     }
+
+    &.show-all-holes {
```
broadwell, line 136: This functionality is fine for now, but I can imagine that the PIs would prefer that these highlights include a rectangle that delineates exactly the start/stop ticks (as translated into roll pixels) of the corresponding MIDI event -- which would show, for example, the effect of applying the tracker bar width extension to each hole group.
Supporting the above would involve matching each hole grouping from the hole report with its corresponding MIDI event, but this is certainly doable given that we have the midi key assignment for each hole group as well as its start pixel/tick. This cross-reference of the MIDI and hole report also would be needed to provide the note velocity value on the popup info for each note (because velocities are only provided in the MIDI data), which is another feature that would be appreciated, I think.

src/components/RollViewer.svelte
simonwiles: Yeah, you're right.  Turns out it's more complicated to increase the hover target than just this (though I think this stays anyway, unless we drop this feature altogether).  Reckon it's worth it to pursue a bigger hover target?

src/components/RollViewer.svelte
simonwiles: This is a different feature, then, I suppose.  What do you think -- scupper this PR and put the feature you describe on the list for future consideration?

src/components/AdvancedSettings.svelte
```diff
@@ -81,9 +81,13 @@
   <fieldset>
     <legend>Visualization Settings</legend>
     <div>
-      Show details for Active Notes:
+      Show Details for Active Notes:
       <input type="checkbox" bind:checked={$userSettings.activeNoteDetails} />
     </div>
+    <div>
+      Highlight All Holes:
```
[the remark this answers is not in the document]
simonwiles: "Highlight All Holes" is certainly a shorthand.  I'm not sure the behaviour is ambiguous, though -- you can immediately see what it does.  Happy to go with whatever verbiage you prefer, if this PR is to be merged at all.

src/components/RollViewer.svelte
broadwell: I'm torn, because I like the green semi-transparent highlighting effect here and think that it does help one to notice hole grouping errors quickly while skimming through the roll, which is certainly a feature the PIs want. But I also think that the viz described above is kind of a superset of this, and so perhaps it's best to keep this PR in reserve until we investigate the advanced MIDI event viz further.

src/components/RollViewer.svelte
broadwell: Providing a bigger hover target would improve the user experience noticeably, I think. Maybe it's worth a separate PR?

src/components/AdvancedSettings.svelte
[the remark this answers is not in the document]
broadwell: Yes the behavior isn't ambiguous, and if the user is unsure about the meaning, all they have to do is click on the checkbox to be enlightened. But it would be nice if we could find some verbiage that would make the checkbox's function a bit more clear before the click. It's not a high priority for now, though.

simonwiles:
Closing in favour of #79.

PR #56, opened 2021-06-23
Pedaling Map

simonwiles:
This doesn't add much in terms of visible user-facing functionality except:
1) you can watch the pedals go up and down as you scroll a non-playing roll;
2) you can re-enable roll pedaling in the middle of a roll and have the pedals activated as appropriate without having to wait for the next relevant event.

I think that's basically it.  But this was needed, and paves the way for the same thing to be done with the activeKeys store so that the keyboard behaves the same way.


broadwell:
This works fine for me, and the code certainly looks as though it is preparing the way for a similar behavior with the key event "map."

PR #57, opened 2021-06-26
Fix upward, downward panning limits for upward scrolling rolls

broadwell:
For the 88-note rolls, which recently received better playback/note highlighting support, the current panning constraints disallow upward panning via the up/down panning buttons or <kbd>ctrl</kbd>-mousewheel at a short, seemingly arbitrary distance up the roll (the location actually is related to the difference between the image length and the pixel location of the first "tick") and allow downward panning to way beyond the bottom (start) of the roll. This adjustment to the clamping limits fixes both issues.

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -347,8 +347,8 @@
           )
         : clamp(
             firstHolePx - centerY - delta,
-            -firstHolePx,
-            imageLength - firstHolePx,
+            firstHolePx - imageLength,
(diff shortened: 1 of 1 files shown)


simonwiles (approved):

simonwiles:
Awesome, thanks.  I'd noticed this but hadn't got round to doing anything about it.  For a three- (or really one-) line fix, I imagine it would have taken me a while to wrap my head around it properly and get it right.  Thanks!

PR #58, opened 2021-07-01
Tooling Updates

simonwiles:
I have (temporarily?) abandoned the SvelteKit conversion due to some funkiness around the way `@tonejs/piano` handles its CJS/ESM imports (esp. w/r/t Typescript interfaces) which make it impossible to compile for SvelteKit-style SSR at present.

This PR contains a some cherry-picked tooling updates (and necessitated code updates) that were performed as part of that effort and which are desirable in any event.


broadwell:
Svelte's shifting goalposts for the `:global` styles is annoying, but I'm also inclined to let it go, as long as everything works now (which it does).
The code streamlining/linting looks good!

PR #59, opened 2021-07-01
`<RollSelector/>` searches are accent insensitive

simonwiles:
There's more to come here still, but I didn't want to wait to get your eyes on this.

PR also includes a fix for the dropdown indicator getting left behind when the left panel is collapsed.

### DIFF
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -97,8 +97,26 @@
   let input;
   let list;
 
+  const unDecomposableMap = {
+    Ł: "L",
+    ł: "l",
+    ß: "ss",
+    Æ: "AE",
+    æ: "ae",
+    Ø: "O",
+    ø: "o",
+  };
+
+  const unDecomposableRegex = new RegExp(
+    Object.keys(unDecomposableMap).join("|"),
+    "g",
+  );
+
   const stripDiacritics = (str) =>
-    str.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
+    str
+      .replace(unDecomposableRegex, (m) => unDecomposableMap[m])
+      .normalize("NFD")
+      .replace(/[\u0300-\u036f]/g, "");
 
   const markupMatches = (label, searchContent, searchParts) => {
     const matchExtents = [];
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -102,14 +102,24 @@
 
   const markupMatches = (label, searchContent, searchParts) => {
     const indices = [];
+    const mergedIndices = [];
     let markedUp = label;
+
     searchParts.forEach((searchPart) => {
       let idx = -1;
       while ((idx = searchContent.indexOf(searchPart, idx + 1)) > -1)
         indices.push([idx, idx + searchPart.length]);
     });
 
-    indices
+    indices.sort().forEach(([start, end], i) => {
+      if (mergedIndices[i - 1] && mergedIndices[i - 1][1] > start) {
+        mergedIndices[i - 1][1] = Math.max(mergedIndices[i - 1][1], end);
+      } else {
+        mergedIndices.push([start, end]);
+      }
+    });
+
+    mergedIndices
       .sort()
       .reverse()
       .forEach(([start, end]) => {
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -98,12 +98,9 @@
   let list;
 
   const unDecomposableMap = {
-    Ł: "L",
     ł: "l",
     ß: "ss",
-    Æ: "AE",
     æ: "ae",
-    Ø: "O",
     ø: "o",
   };
 
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -20,7 +20,6 @@
       transform: rotate(-45deg);
       transform-origin: center;
       width: 0.625em;
-      z-index: 4;
     }
   }
 
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -98,6 +98,9 @@
   let input;
   let list;
 
+  const stripDiacritics = (str) =>
+    str.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
+
   const selectListItem = (
     listItem = filteredListItems[activeListItemIndex],
   ) => {
@@ -139,7 +142,7 @@
     activeListItemIndex = 0;
 
     if (!text) return;
-    const filteredText = text
+    const filteredText = stripDiacritics(text)
       .replace(/[&/\\#,+()$~%.'":*?<>{}]/g, " ")
       .trim()
       .toLowerCase();
@@ -169,9 +172,9 @@
 
   const prepareListItems = () => {
     listItems = items.map((item) => ({
-      searchContent: (searchFieldName ? item[searchFieldName] : item)
-        .toLowerCase()
-        .trim(),
+      searchContent: stripDiacritics(
+        (searchFieldName ? item[searchFieldName] : item).toLowerCase().trim(),
+      ),
       label: labelFieldName ? item[labelFieldName] : item,
       item,
     }));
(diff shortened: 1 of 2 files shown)


src/ui-components/FilteredSelect.svelte
```diff
@@ -98,6 +97,27 @@
   let input;
   let list;
 
+  const stripDiacritics = (str) =>
+    str.normalize("NFD").replace(/[\u0300-\u036f]/g, "");
```
broadwell, line 101: Do you think it's worth normalizing/ASCII-ifying non-diacritic, non-decomposable characters? The only ones I can think of that might be worth addressing in this Western/European context are ø → o and ß → ss -- though these would be quite rare, assuming our catalogers haven't manually normalized them already.

broadwell (approved):
This is great! I hadn't even noticed that the drop-down was prone to sticking around when the left pane was collapsed.
I'd be interested to get your feedback on my code comment, but am happy to merge this as is.

src/ui-components/FilteredSelect.svelte
simonwiles: Yeah, we certainly could, if you think the specifics of this data warrant it.  It all gets a bit ad-hoc when we start using manually (non-Unicode) specified compatibility forms.  The "ß" is in a kind of third state, because `"ß".toUpperCase().toLowerCase() === "ss"`, which has always struck me as a bit weird.  We don't seem to have any of either of these in `catalog.json` at present, so it would a future-proofing thing?

Also worth noting that the way this works now you *can't* distinguish accented and unaccented forms -- so if you *want* to search for "espanñola" and exclude "espanola" that's not possible.

broadwell: Not being able to distinguish the accented forms is OK with me, and seems to be pretty typical practice even in international contexts, though SUPRA's European collaborators might grumble a bit.

This definitely would be a future-proofing step at this point given that none of the non-decomposable characters are in the data set at present, which makes it even more ad-hoc and speculative because we have to try to guess which alphabets are likely to be turn up in a larger roll set and also not be not normalized by the catalogers (note that the other, less "official" fields of the MODS files for the current set of rolls already do tend to contain a much wider variety of spellings and characters, so some normalization actually occurs between the MODS fields).

Anyway, the more challenging scripts I think we'd be likely to see in a corpus of piano music are German, Dano-Norwegian, and Polish/West Slavic. So the mappings to add would be:
Polish:
Ł -> L
ł -> l
German:
ß -> ss
Dano-Norwegian:
Æ -> AE
æ -> ae
Ø -> o
ø -> o

It's pretty much a given that I'll turn out to have missed something, though...

simonwiles: > Not being able to distinguish the accented forms is OK with me, and seems to be pretty typical practice even in international contexts, though SUPRA's European collaborators might grumble a bit.

There's also the issue of whether the metadata is actually consistent and correct about these things in the first place.

> (note that the other, less "official" fields of the MODS files for the current set of rolls already do tend to contain a much wider variety of spellings and characters, so some normalization actually occurs between the MODS fields).

Yeah there seems to be some funky stuff -- did you notice "Li͡apunov" with a combining double breve!?!

> So the mappings to add would be:
> ...
>
> It's pretty much a given that I'll turn out to have missed something, though...

Indeed -- I'll add these in, since you've gone to the trouble of listing them out, then, and we'll leave it at that?

simonwiles: Ø -> **O**, of course.

simonwiles:
Okay, there are updates:
1. to fix cases where matches weren't in l-t-r order, and the marking-up got borked;
2. to fix cases where matches overlapped, and the marking-up got borked; and
3. to add a manual compatibility substitution map as per our discussion.

src/ui-components/FilteredSelect.svelte
broadwell: Right -- though if the matching is case-insensitive (i.e., against the lowercase version of the ASCII-ified search string, I suppose this wouldn't make a difference.

broadwell:
I verified the substitution map by adding a bogus catalog entry with some of those characters. It worked fine!
Glad you thought to check whether typing the search terms out of order (compared to how they appear on the target) works properly with the highlighting/autocomplete. Now that you've brought it to my attention, though, I've noticed that even the latest update isn't entirely bulletproof, e.g.,
Screen Shot 2021-07-02 at 12 57 26 AM
Screen Shot 2021-07-02 at 12 55 16 AM
Do we care that much? These are pretty contrived searches, but maybe there's a quick fix for this?

src/ui-components/FilteredSelect.svelte
simonwiles: Ah, yes, good point... I probably shouldn't have included the uppercase versions in the map at all...

simonwiles:
> Do we care that much? These are pretty contrived searches, but maybe there's a quick fix for this?

Hmm, yeah, I'll fix this (I thought I had).  One cheap solution is to prevent filtering on single characters, but I didn't love the UX, and I'd thought it wasn't necessary.  I'll take a look this morning.

simonwiles:
Yeah, as suspected, this was a bug in my algorithm.  This is a famously hard problem without a straightforward known-best solution for all cases.  In this case, certain allowances are made (case- and accent- insensitive search only, for example) which serve to simplify the problem and, hopefully, allow a relatively simple "bullet-proof" implementation -- I just made an implementation mistake :)  Thanks for catching it!

I've also removed the uppercase variants of the additional normalizable characters you identified, and refactored the text normalization a little (mainly to ensure that strings are lowercased before the substitutions are performed).

simonwiles:
> I verified the substitution map by adding a bogus catalog entry with some of those characters. It worked fine!

Unfortunately, it doesn't work properly with the Æ ligature and the eszett.  It will need some exception code to deal with these, which will introduce an extra order of complexity :(

src/ui-components/FilteredSelect.svelte
broadwell: Also yeah, Li͡apunov was a real surprise. Good on SearchWorks for not mangling it, I guess. I wasn't familiar with the name, and the breve is also used in music notation, so I was vaguely concerned that perhaps one had gotten loose in some music cataloging datastream and come to rest there.

broadwell:
> Unfortunately, it doesn't work properly with the Æ ligature and the eszett. It will need some exception code to deal with these, which will introduce an extra order of complexity :(
Ok yeah, I didn't test those... I'd say that if the extra complexity is at all significant, I'd be fine with leaving them unmapped.
I assume it's not feasible/worthwhile just to punt all of this over to the unidecode library?

simonwiles:
> I assume it's not feasible/worthwhile just to punt all of this over to the unidecode library?

Wouldn't help enormously, I don't think -- the normalization and matching is the easy bit.  As usual, it's the markup-up of the matched results that requires contortions.

src/ui-components/FilteredSelect.svelte
simonwiles: I guess it's just a fringe romanization of the Cyrillic -- but of course when you're cataloging sometimes you need to use the form used on the artefact itself, and you can't "update" things to modern standardization.  Had lots of fun with historical (and sometimes uniquely idiosyncratic *hapax legomena*) romanizations of East Asian stuff in the past, as you can probably imagine.

simonwiles:
Well this was predictably involved, but I *think* this is solid now.  It certainly answers all the cases we've tried so far, which exceeds the requirements of the current corpus by some distance.

broadwell:
Here are a couple of contrived searches that still cause trouble for the highlighting. Do you think it's worth tracking these down? Maybe just call it good enough for now? It certainly works fine for most plausible searches I can think of... Or, if the class of searches that cause problems with the highlighting is well defined, maybe just not do the highlighting for those cases?
Screen Shot 2021-07-06 at 3 53 59 PM
Screen Shot 2021-07-06 at 3 54 54 PM

simonwiles:
Thanks, this was another bug in the algorithm.  This component could really use a test suite...  This case is straightforward enough, anyway, even if it did take a little head scratching.

broadwell:
OK great, I'm not seeing any of the fragmented markup in the search results/highlights anymore! One last issue, though -- there's an obvious namespace collision between the search text highlighting and the hole highlighting, which I assume isn't intentional or desired:
Screen Shot 2021-07-06 at 6 49 05 PM

simonwiles:
Yep, I have a commit that fixes this in the next PR, that covers rich formatting of items in the selector along with a couple of other small things.  Faceting is then a subsequent PR, because I think the intermediate step will require some discussion and probably some back-and-forth too.

Thanks for sticking with this one, sorry it was such a pain -- I knew the marking up of the "search results" was going to be troublesome, it always is....

PR #60, opened 2021-07-07
`<RollSelector/>` rich formatting and adding label numbers

simonwiles:
This PR fixes a couple of incidental issues (the first two commits), and then adds some rich styling options to the dropdown list items and the input field itself.  It does this by proving a `postMarkup()` hook that allows parent components to alter the display of the various elements *after* they've been filtered and marked up for matches.  In order to make this possible for the input box too, the `<input type="text"/>` element has been replaced with a `<span contenteditable="true"/>`.

This gives lots of flexibility, and from a purely aesthetic point of view I like the way it looks now.  However, the way I've split up the label number is probably controversial/flat-out wrong.  The numbers don't appear to conflict at all, but I assume that's just a fluke?  Anyway, I assume we'll probably need to adjust this somehow (or do it in a slightly different way altogether).  Do you have a clear steer to offer on this, or does this need to be a conversation involving the PIs?

Edit: There are also a couple where the `label` field in the catalogue has only a number, which currently results in an empty `[]` at the end of the line.  In the unlikely event that the current arrangement is acceptable, I'll fix that.

### DIFF
--- a/src/components/RollSelector.svelte
+++ b/src/components/RollSelector.svelte
@@ -1,10 +1,13 @@
-<style>
+<style lang="scss">
   :global(small) {
     color: grey;
     display: inline-block;
-    width: 5ch;
     text-align: right;
     margin-right: 1ch;
+
+    &:first-child {
+      width: 5ch;
+    }
   }
 </style>
 
--- a/src/components/RollSelector.svelte
+++ b/src/components/RollSelector.svelte
@@ -1,5 +1,5 @@
 <style lang="scss">
-  :global(small) {
+  :global(.filtered-select small) {
     color: grey;
     display: inline-block;
     text-align: right;
--- a/src/components/RollSelector.svelte
+++ b/src/components/RollSelector.svelte
@@ -1,6 +1,10 @@
 <style>
   :global(small) {
     color: grey;
+    display: inline-block;
+    width: 5ch;
+    text-align: right;
+    margin-right: 1ch;
   }
 </style>
 
@@ -10,7 +14,10 @@
 
   const listItems = catalog.map((item) => ({
     ...item,
-    _label: `[${item.label}] ${item.title}`,
+    _label: `${item.label.match(/^\d+/)} ${item.title} [${item.label.replace(
+      /^\d*\s?/,
+      "",
+    )}]`,
   }));
   export let currentRoll =
     listItems[Math.floor(Math.random() * catalog.length)];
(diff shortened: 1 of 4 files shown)


broadwell:
The rich formatting looks great! I do think the PIs probably will prefer the label number and company name to appear together, perhaps to the right of the other info, and still in square brackets, e.g., `[Welte-Mignon 520]`. but I'd be happy to merge this as is and then ask them for feedback.
First, though, I've noticed another corner case with the search field (also present in the previous PR): typing a bunch of spaces or other punctuation after an initial letter can quickly trigger memory exhaustion. Do any mitigation strategies come to mind? It's not something anyone would type intentionally, but it does highlight the potential that a sticky key could end up crashing the page...
Screen Shot 2021-07-07 at 1 00 56 PM
Screen Shot 2021-07-07 at 1 51 08 PM
Screen Shot 2021-07-07 at 12 46 16 PM

simonwiles:
Yep, thanks.  I can reproduce but only when I add another (non-whitespace, non-punctuation) character *after* the long string of spaces/punctuation (and understand this and have a fix for this ready) -- is that the same for you?

broadwell:
> Yep, thanks. I can reproduce but only when I add another (non-whitespace, non-punctuation) character _after_ the long string of spaces/punctuation (and understand this and have a fix for this ready) -- is that the same for you?

OK yeah, I forgot to include that detail: the warning only appears when I try to type another character after what you see in the screenshots (but the character doesn't appear, due to the resource exhaustion).

simonwiles:
Okay, great, thanks -- fix(es) committed.

The algorithm involves `searchString.split(" ")` and 
```
"a              b".split(" ") === ["a", "", "", "", "", "", "", "", "", "", "", "", "", "", "b"]
```
(and punctuation is replaced with white space, hence the same behaviour with `a...................b`).

The first extra commit collapses consecutive whitespace (after the punctuation substitution), and the second enforces an arbitrary limit of 8 search terms, with any excess being silently discarded.

broadwell:
Cool -- I was thinking that an arbitrary search term limit might be called for here.
Regarding the formatting: we'll probably want to add some logic to handle empty values, see e.g., this 88-note roll with an issue number but no company name:
Screen Shot 2021-07-07 at 3 46 59 PM
Probably this can wait until we get feedback about the desired format of the dropdown items.

simonwiles:
Yep, agreed -- see the note in my original "post".  I think it probably makes sense to merge this first, though, and deal with changes later, if that's cool with you?  (want to get the faceting in, too)

broadwell:
OK right, I see the relevant section in your comment! Let's merge this now and deal with the empty values when we have a better handle on the desired format.

simonwiles:
Thanks -- like you I have a feeling we'll be doing something different with this in the final analysis, so can't be bothered to write code I'm fairly sure will be thrown away :)

PR #61, opened 2021-07-07
`<RollSelector/>` Faceting

simonwiles:
This PR adds faceting on roll type.  The UI is plain, but servicable, I think?  I've not found any bugs, but you usually manage to find something (thank you)!

### DIFF
--- a/src/components/RollSelector.svelte
+++ b/src/components/RollSelector.svelte
@@ -31,5 +31,6 @@
   bind:selectedItem={currentRoll}
   labelFieldName="_label"
   searchFieldName="_label"
+  facetFieldName="type"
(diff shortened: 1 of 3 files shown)


broadwell:
I think the UI looks fine, and I couldn't scare up any bugs at all.
I'm not a huge fan of "Filtered:", but I agree that some explanatory text needs to be there, and can't think of a word that would be an obviously better fit.
Also, it's possible that the preferred faceting mode is for the pills to be multi-select rather than exclusive radio buttons, but this preference may not become clear until there are more roll types to choose from. Let's go with the current behavior for now.

simonwiles:
Yeah, I agree with those points, but this is something to iterate from -- I felt like multi-select was overkill at this point, but of course it's easy enough to add.  I'm trying to practice just a little bit of YAGNI :)

PR #62, opened 2021-07-08
Misc. Updates

simonwiles:
This is a collection of miscellaneous updates that I would like to get in before tomorrow's meeting.

* The `<RollSelector/>` dropdown should close if you click on the input field while it's open.
* `build/**` should be excluded from the prettier check in `yarn lint`.
* I've tried forcing the fixed-width font for `<kbd/>` elements (i.e. the keyboard shortcut indicators).  I noticed the look badly out of alignment on Macs, and suppose this is the result of different systems selecting a different typeface for `monospace`, so I've tried forcing it to use "Courier New" on all systems.
* "Stop" is changed to "Rewind" -- temporary thing, but the commit's been sitting on a random local branch for weeks, so it might as well go in here.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 node_modules/**
 src/assets/**
+build/**
--- a/src/components/PlaybackControls.svelte
+++ b/src/components/PlaybackControls.svelte
@@ -39,7 +39,7 @@
 
 <div id="playback-controls">
   <button type="button" on:click={playPauseApp}>Play/Pause</button>
-  <button type="button" on:click={stopApp}>Stop</button>
+  <button type="button" on:click={stopApp}>Rewind</button>
   <button
     type="button"
     class:pedal-on={$softOnOff}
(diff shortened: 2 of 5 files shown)


simonwiles:
Snuck in an extra commit for the one label number with a period in it.

broadwell:
Excellent!
> The `<RollSelector/>` dropdown should close if you click on the input field while it's open.  

I was thinking this would be nice, but forgot to ask about it.

> I've tried forcing the fixed-width font for `<kbd/>` elements  

The key hints now look a lot better on Firefox, Chrome/Opera, and Safari on Mac.

simonwiles:
> > The `<RollSelector/>` dropdown should close if you click on the input field while it's open.
> 
> I was thinking this would be nice, but forgot to ask about it.

Yeah, I wanted this in the first PR, but it was annoyingly tricky to get right, so I just left it and moved on.  Came back to it a couple of times before I got it right.  Arguably it's not 100% perfect now, because if you try to reposition the cursor using the mouse it closes the dropdown, but in general I'm happy with the UX for the dropdown, so I'm ready to leave it for a while now :)

PR #63, opened 2021-07-21
DRUID as URL parameter

simonwiles:
I think this should cut it for now, at least.  I had to rejig the way (or rather where) the munging of the catalog JSON occurs, but otherwise this is fairly clean.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -76,6 +76,8 @@
   import Notification, { notify } from "./ui-components/Notification.svelte";
   import FlexCollapsible from "./ui-components/FlexCollapsible.svelte";
 
+  import catalog from "./assets/catalog.json";
+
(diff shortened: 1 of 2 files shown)


simonwiles:
Decided this needs a little finessing -- more commits incoming.

simonwiles:
Change of plan -- my intention is to introduce some handling for malformed or just incorrect DRUIDs, but this will need some only-partially-related fixes to go in first.  I'm going to leave this as-is, and ask to you review (and hopefully merge) it like this, and address the changes I wanted to make in a subsequent PR.  Sorry for messing you around!

broadwell:
This works fine for me as a basic level of functionality, so I'll go ahead and merge it, thanks! Happy to review the refinements in a separate PR.

PR #64, opened 2021-07-21
Misc. Fixes and Updates

simonwiles:
The Aristotelian _teloi_ for these changes are:
1. To display a notification if a malformed or otherwise unavailable DRUID is provided in the querystring.
2. To have the app behave more like a proper SPA when updating the querystring (and therefore the URL) either from the dropdown or using the browser back/forward buttons.

To achieve the former, it was necessary to allow for the creation of "unclosable" notifications, which in turn necessitated some updates to handling of `window.onClick` events in the `<RollSelector/>` and `<RollViewer/`> components.  To allow the latter, some additional routing logic and fallbacks were required.

There are two (well, at least two ;)) possibly problematic parts of this PR that I'd like to draw attention to:
1. To close the `<RollSelector/>` dropdown when "clicking outside", there is a `window.onClick` event that tests if the click is outside the roll selector component.  Previously this had a check for `event.defaultPrevented`, but this was causing it not to fire when clicking on the roll viewer (I think OSD calls `event.preventDefault()` on everything).  While not strictly necessary to accommodate (1) above, I've removed this check for a better UX.  Tbh, however, I'm not sure why it was there in the first place, so it's possible some other part of the UX is degraded by removing it.  I couldn't find (or think of) anything, but if you come across anything funky, it might be to do with that.
2. Connected with the preceding point, clicking on the roll image when the dropdown is open closes the dropdown as expected, but was also causing OSD to zoom in, which was icky.  After some brief consideration, I simply disabled `clickToZoom` on OSD.  This is perhaps a bit heavy-handed, but the alternatives are quite involved, and for my part at least I like it better with `clickToZoom` off.  There are plenty of other ways to zoom in (`clickToZoom` only ever zoomed *in* anyway).  However, I appreciate this may be controversial and need to be reverted.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -199,24 +199,13 @@
         if (params.has("druid") && params.get("druid") !== currentRoll.druid) {
           const url = new URL(window.location);
           url.searchParams.set("druid", currentRoll.druid);
-          window.history.pushState({}, "", url);
+          window.history.pushState({ roll: currentRoll }, "", url);
         }
       },
     );
   };
 
-  onMount(async () => {
-    ({
-      midiSamplePlayer,
-      pianoReady,
-      updatePlayer,
-      startNote,
-      stopNote,
-      pausePlayback,
-      startPlayback,
-      resetPlayback,
-    } = samplePlayer);
-
+  const setCurrentRollFromUrl = () => {
     const params = new URLSearchParams(window.location.search);
     if (params.has("druid")) {
       const druid = params.get("druid");
@@ -236,6 +225,21 @@
       currentRoll =
         rollListItems[Math.floor(Math.random() * rollListItems.length)];
     }
+  };
+
+  onMount(async () => {
+    ({
+      midiSamplePlayer,
+      pianoReady,
+      updatePlayer,
+      startNote,
+      stopNote,
+      pausePlayback,
+      startPlayback,
+      resetPlayback,
+    } = samplePlayer);
+
+    setCurrentRollFromUrl();
   });
 
   $: if (currentRoll !== previousRoll) loadRoll(currentRoll);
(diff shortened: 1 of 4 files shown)


broadwell:
This all works well for me. I haven't been able to find any UI flakiness that could be attributed to concern 1) above. Regarding concern 2), I actually found the roll click-to-zoom behavior to be more confusing that helpful, so I'm not sad to see it go, especially considering that mousewheel zoom works so well, both in and out (as will pinch zoom, eventually).

Just a comment: I do find it slightly inconsistent that the DRUID for a roll is *not* added to the URL query string until/unless the query string feature actually is used to load a roll (i.e., by typing/pasting the augmented URL into the address bar), after which it remains in the address bar and is updated appropriately when a new roll is selected, *until* it is manually deleted from the address bar, after which it will not appear again (unless on is re-added manually). I'd think that it would be preferable for the DRUID to be present in the URL query string always, even when the app is loaded to a random roll, especially as the presence of the DRUID in the URL provides a rudimentary bookmarking feature. On the other hand, all of this URL-based functionality is likely to change when more sophisticated routing is added, so it's not worth deliberating over at the moment.

simonwiles:
> Regarding concern 2), I actually found the roll click-to-zoom behavior to be more confusing that helpful, so I'm not sad to see it go, especially considering that mousewheel zoom works so well, both in and out (as will pinch zoom, eventually).

Yeah, that was definitely my feeling too, tbh.  Let's see if anyone else notices/complains? ;)

> Just a comment: I do find it slightly inconsistent that the DRUID for a roll is not added to the URL query string until/unless the query string feature actually is used to load a roll (i.e., by typing/pasting the augmented URL into the address bar), after which it remains in the address bar and is updated appropriately when a new roll is selected, until it is manually deleted from the address bar, after which it will not appear again (unless on is re-added manually). I'd think that it would be preferable for the DRUID to be present in the URL query string always, even when the app is loaded to a random roll, especially as the presence of the DRUID in the URL provides a rudimentary bookmarking feature. On the other hand, all of this URL-based functionality is likely to change when more sophisticated routing is added, so it's not worth deliberating over at the moment.

That's a fair point.  I did think about this, and decided that I thought it was preferable for the app to operate as previously unless the URL querystring param "mode" was invoked.  Reading your comment and thinking about it again, I think you've probably changed my mind.  Probably wait to update it, though.

PR #65, opened 2021-07-23
Fix emulated tempo bug

simonwiles:
This PR is to address the issue with toggling roll acceleration emulation not working properly.  The issue was that the `"Set tempo"` events that `midiplayer-js` fires occur **after** the tempo has already been altered.  This means that a) there's no need to listen for these events and manually adjust the tempo in response to them; and b) ignoring these events when `$useMidiTempoEventsOnOff === false` has no effect.  Fortunately our MIDI files have the tempo events on the first MIDI track which otherwise contains only metadata (i.e. no notes) -- so we can simply toggle this track to enable/disable tempo events from the MIDI with no side-effects.

After a lot of testing and re-testing, I just removed dozens of lines of debugging code from this branch and it boils down to three lines in and three lines out :)

### DIFF
--- a/src/components/SamplePlayer.svelte
+++ b/src/components/SamplePlayer.svelte
@@ -201,7 +201,7 @@
 
   midiSamplePlayer.on(
     "midiEvent",
-    ({ name, value, number, noteNumber, velocity, data }) => {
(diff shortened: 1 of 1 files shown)


broadwell:
This works fine for me!

PR #66, opened 2021-07-23
Roll Panning Dynamics and a fix for redundant brackets

simonwiles:
Dual PR (sorry!  -- commits are clean!):

* Drops brackets when a label number is available but no label name (though presumably there's a cataloging error here?)
* Adjusts the dynamics of the vertical panning buttons to facilitate fine-grained (< hole-by-hole) panning while still allowing their use for navigation.

I'd fiddled with the latter and arrived at these values, but am open to feedback and further suggestions for what (else?) to do with this.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -96,13 +96,13 @@
   let startPlayback;
   let resetPlayback;
 
-  const rollListItems = catalog.map((item) => ({
-    ...item,
-    _label: `${item.label.match(/^[\d.]+/)} ${item.title} [${item.label.replace(
-      /^[\d.]+\s?/,
-      "",
-    )}]`,
-  }));
+  const rollListItems = catalog.map((item) => {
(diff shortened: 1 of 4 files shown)


simonwiles:
Two additional commits to fix issues I'd overlooked.

broadwell:
> Drops brackets when a label number is available but no label name (though presumably there's a cataloging error here?)

For [this one](https://searchworks.stanford.edu/view/13406490), it seems that they legitimately just weren't able to find a definitive indication of the label, and so left it blank.

> Adjusts the dynamics of the vertical panning buttons to facilitate fine-grained (< hole-by-hole) panning while still allowing their use for navigation.

The `easingInterval` is great! It might be nice to have similar behavior for the +/- zoom buttons, or at least to be able to zoom repeatedly by holding the button down.

PR #67, opened 2021-07-23
Piano keyboard highlighting when panning / navigating

simonwiles:
Okay, now that the improved panning behaviour is in, we can do the key highlighting as we pan / step-through.  This also sets the highlighted keys when jumping / navigating otherwise.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -130,7 +130,6 @@
   const skipToTick = (tick) => {
     if (tick < 0) pausePlayback();
     $currentTick = tick;
-    activeNotes.reset();
     updatePlayer(() => midiSamplePlayer.skipToTick($currentTick));
   };
 
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -143,15 +143,12 @@
   import { fade } from "svelte/transition";
   import OpenSeadragon from "openseadragon";
   import { rollMetadata, currentTick, userSettings } from "../stores";
-  import { clamp } from "../utils";
+  import { clamp, getNoteName } from "../utils";
   import RollViewerControls from "./RollViewerControls.svelte";
 
   export let imageUrl;
   export let holesByTickInterval;
   export let skipToTick;
-
-  const WELTE_RED_FIRST_NOTE = 24;
-  const WELTE_RED_LAST_NOTE = 103;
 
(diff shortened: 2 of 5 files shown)


broadwell:
These are great additions, and everything works fine for me.

Just noting here a couple of weird things that can happen with the keyboard/piano when panning or during playback -- these were both present before this PR:
- Some keys can remain stuck "on" when playback reaches the end of the roll. This is much more likely to happen with the 88-note rolls (e.g., https://sul-cidr.github.io/pianolatron/?druid=pz594dj8436) because the reproducing rolls usually have some "margin" after the last note due to the control holes. It's also possible that the Expressionizer adds some padding to the MIDI file as well. These stuck keys are cleared if the "Rewind" button is pressed, but otherwise can persist through panning or restarting via the "Play/Pause" button. It's not clear if the best fix would involve changing the player's behavior (e.g., continuing for a few more ticks after the last event) or modifying the MIDI files.
- Hole highlighting jumps ahead of the tracker line when panning at full speed. This is definitely a low-priority concern. Addressing it probably would require modifying the synchronization with the OSD viewer, which might break other things.

simonwiles:
Thanks!

> Some keys can remain stuck "on" when playback reaches the end of the roll. This is much more likely to happen with the 88-note rolls (e.g., https://sul-cidr.github.io/pianolatron/?druid=pz594dj8436) because the reproducing rolls usually have some "margin" after the last note due to the control holes. It's also possible that the Expressionizer adds some padding to the MIDI file as well. These stuck keys are cleared if the "Rewind" button is pressed, but otherwise can persist through panning or restarting via the "Play/Pause" button. It's not clear if the best fix would involve changing the player's behavior (e.g., continuing for a few more ticks after the last event) or modifying the MIDI files.

This should be fairly easy to address I think.

> Hole highlighting jumps ahead of the tracker line when panning at full speed. This is definitely a low-priority concern. Addressing it probably would require modifying the synchronization with the OSD viewer, which might break other things.

I've notice this too, and I think it's just because OSD take a while to "catch up" because it animates its way to the target position while the playback and associated hole highlight jumps right there.  It's one of those things where it might be a fairly simple problem but I've a nagging suspicion that it might be a can of worms and I'm not sure that the behaviour is egregious as it stands, so I've left the can close for now :)  Will take a look at it and see.

PR #68, opened 2021-07-25
Mini-map / Navigator ?

simonwiles:
This is a draft PR just to invite you to take a look at this.  If we go this way we get this much this easily (although there's still a bit of work to do, of course).  On the other hand, now that I see it, I'm not sure there's much use in the "mini-map", given that you can't see anything at all...  I also experimented with just having a vertical slider, and am considering a hand-rolled solution with one rectangle (`<div/>`) inside another where the former is sized and moves to reflect what is shown in the viewport w/r/t the whole image.

Thoughts and opinions?

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -63,6 +63,7 @@
     animatePan,
     playExpressionsOnOff,
     rollPedalingOnOff,
+    playbackProgress,
   } from "../stores";
   import {
     clamp,
@@ -157,6 +158,7 @@
       }
     });
   };
+  let osdNavDisplayRegion;
 
   const createMark = (hole) => {
     const {
@@ -270,6 +272,7 @@
     );
 
     viewport.panTo(lineCenter, !$animatePan);
+    osdNavDisplayRegion.dataset.label = ($playbackProgress * 100).toFixed(1);
   };
 
   const highlightHoles = (tick) => {
@@ -315,11 +318,14 @@
 
     ({ viewport } = openSeadragon);
 
+    osdNavDisplayRegion = openSeadragon.navigator.displayRegion;
+
     // Override some styles that OSD sets directly on the elements
     openSeadragon.navigator.element.style.border = "none";
     openSeadragon.navigator.element.parentElement.style.backgroundColor =
       "#666";
     openSeadragon.navigator.displayRegion.style.display = "block";
+    openSeadragon.navigator.displayRegion.style.overflow = "visible";
     openSeadragon.navigator.displayRegion.style.left = "0";
     openSeadragon.navigator.displayRegion.style.width = "100%";
 
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -357,15 +357,15 @@
       createHolesOverlaySvg();
       advanceToTick(0);
     });
-    openSeadragon.addHandler("canvas-drag", () => {
+    openSeadragon.addHandler("canvas-drag", () => (strafing = true));
+    openSeadragon.addHandler("canvas-drag-end", () => (strafing = false));
+    openSeadragon.addHandler("pan", () => {
       const viewportCenter = viewport.getCenter(false);
       const imgCenter = viewport.viewportToImageCoordinates(viewportCenter);
       skipToTick(
         scrollDownwards ? imgCenter.y - firstHolePx : firstHolePx - imgCenter.y,
       );
-      strafing = true;
     });
-    openSeadragon.addHandler("canvas-drag-end", () => (strafing = false));
     openSeadragon.addHandler("open", () => {
       const tiledImage = viewport.viewer.world.getItemAt(0);
       tiledImage.addOnceHandler(
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -310,9 +310,44 @@
       navigatorLeft: "calc(100% - 40px)",
       navigatorHeight: "100%",
       navigatorWidth: "40px",
+      navigatorDisplayRegionColor: "white",
     });
 
     ({ viewport } = openSeadragon);
+
+    openSeadragon.navigator.update = function navUpdate(mainViewport) {
+      // reimplemented based on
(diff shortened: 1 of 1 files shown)


broadwell:
I'd be in favor of showing this to the PIs, even in its draft-y state. It's certainly true that it's difficult to see many salient features on the roll when it's zoomed all the way out, though.

A related, minor concern is that I'm not sure where the progress percentage readout would go if we switch to a vertical slider/mini-map. It's definitely worth displaying it somewhere, and maybe incorporating it into the query string, since that's probably the best way to refer to positions along the roll. Feet and/or pixels are options, too...

simonwiles:
After trying **_many_** different approaches, this one (5ff31f5) has a lot of potential.  I really want to keep the OSD `pan` event listener, as it has so many advantages (not least the fact that it keeps the roll in sync with the app playback position when zooming).

broadwell:
Avoiding the per-tick pan events definitely solved the egregious performance issues previously observed with browsers on MacOS (dropped notes, slowed playback).

Besides some possible styling refinements to the navigator's overlay (mostly involving finessing the location of the number) the only actual bug I've noticed is that the horizontal progress slider seems to default to about 40-50% of the way through any roll on initial load -- even when no roll was loaded previously. The viewer and navigator still jump to the top of the roll on load, and yet the navigator progress readout agrees with the horizontal slider, and the actual playback begins at the midpoint. Presumably this can be fixed by resetting the necessary state values, equivalent to clicking the "Rewind" button.

simonwiles:
>Avoiding the per-tick pan events definitely solved the egregious performance issues previously observed with browsers on MacOS (dropped notes, slowed playback).

Thank goodness...

>Besides some possible styling refinements to the navigator's overlay (mostly involving finessing the location of the number)

The positioning thing was a bug -- worked on Chrom* but not on FF.  Works acceptably on both now (haven't tested Safari...).

> the only actual bug I've noticed is that the horizontal progress slider seems to default to about 40-50% of the way through any roll on initial load -- even when no roll was loaded previously. The viewer and navigator still jump to the top of the roll on load, and yet the navigator progress readout agrees with the horizontal slider, and the actual playback begins at the midpoint. Presumably this can be fixed by resetting the necessary state values, equivalent to clicking the "Rewind" button.

Rewinding the roll was an option, of course, but I'd rather it didn't jump to the wrong position in the first place.  
1bd532f is a bit of a hacky way to achieve this, but short of submitting a feature request / PR against OSD, I think hacky solutions are all we have.  This is my preferred hacky solution :)

There are still some loose ends here -- the panning buttons want moving, we might want to drop the progress slider now, I think I'd like to look again at a label that is also an input field, etc. -- but for my money I think this is ready to go in and go up, at this point.

broadwell:
Yes, the initial positioning is fixed now on all of the Mac OS browser families, even Safari (meaning just that Safari handles playback no worse than before the addition of the navigator).

Another future refinement might be to re-flip the location of the progress label as it nears the beginning or end of the roll -- it's not ideal to have the number readout march off the top of the screen at around 97%. But I agree this is fine to merge as is for now.

simonwiles:
> Another future refinement might be to re-flip the location of the progress label as it nears the beginning or end of the roll -- it's not ideal to have the number readout march off the top of the screen at around 97%. But I agree this is fine to merge as is for now.

How have I not noticed this before -- eek! -- the numbers are on the wrong sides for the rolls that scroll upwards... Fix will come in a new PR shortly.  Thanks...!

broadwell:
Right, yes it only happens for the upwards-scrolling rolls.

A sort of related issue that arises with all of the rolls is that if you zoom out pretty far, then try to pan all the way back so the tracker line is on 0% by clicking and dragging on the main roll viz or the navigator, it won't let you do it -- the top (or bottom) of the navigator box hits the edge of the image and won't go any farther. It is however possible to force it to move beyond the boundaries of the image via the horizontal progress slider or the up/down buttons, but clicking/dragging either image doesn't get you there unless you zoom in so that the boundaries of the navigator box no longer exceed the image edges at 0%.

Literally an edge case, and maybe not the last of its kind...

Screen Shot 2021-08-09 at 11 23 13 PM

simonwiles:
Huh, yeah, I'd not come across that.  You have to be a pretty long way out, of course.  The rewind button still works, too -- likewise playing on if you create the same situation at the other end of the roll.

It's caused by the `constrainDuringPan` option passed to the OSD constructor.  This is something I'd be reluctant to give up, personally -- although after our discussion about horizontal panning today, perhaps I'm the only one who prefers it like this?  Perhaps try setting it to `false` and see what you think?  There doesn't seem to be a simple way to constrain horizontally but not vertically (at first glance, at least), but I seem to be monkey-patching OSD left right and centre at the moment, so this can probably be done (and would afford the best solution in terms of behaviour, imo).

broadwell:
Hmm yeah, I think I do prefer `constrainDuringPan: false` overall, although it does feel quite slippy. More significantly, though, setting it to `false` also gets rid of the tendency of the navigator to go haywire when you try to pan it via click-dragging before a prior zoom animation has completed. Here's an example:
ConstrainPan=true
The main viewer also used to exhibit this kind of behavior, but that seems to have gone away. I think it has something to do with the fact that the zoom animation operates on an exponential scale, while 2D panning is linear. Presumably there's a way to remedy this in the navigator without going to `constrainDuringPan: false`, but I kind of get the sense that OSD is more "comfortable" with that setting in general.

simonwiles:
The odd behaviour is easy enough to address -- I'm about to PR a branch with miscellaneous fixes and updates, so I'll add in a fix for that.

Shall I add `constrainDuringPan: false` too?  I'm a definite -1 -- it makes the panning feel unstable and really very unsatisfactory to me, but I'm ready to be out-voted :)  We could try it and ask the PIs for feedback if you like.

broadwell:
If we can fix the zoom/panning freakouts without using `constrainDuringPan: false` then I'd be fine with that. I do find the panning to be rather twitchy with that setting, which I think is less than desirable especially in the horizontal dimension on the main viewer.

PR #69, opened 2021-07-26
Vertical panning fine-tuning

simonwiles:
I'm pretty sure this won't be able to skip over any note-holes now.  Tried a few approaches, but this felt best.

One side-effect -- this effects the mouse-wheel panning too (well, the previous PR did too, but it's exaggerated now).  The mouse-wheel is fine (good, better probably) for stepping through now, but kinda useless for navigating the roll at this point.  For my part I think I'm probably okay with that, given the upcoming nav / mini-map feature, but not sure if you'll agree?

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -305,7 +305,7 @@
   const panByIncrement = (up = true) => {
     const viewportBounds = viewport.getBounds();
     const imgBounds = viewport.viewportToImageRectangle(viewportBounds);
-    const delta = up ? imgBounds.height / 50 : -imgBounds.height / 50;
+    const delta = up ? imgBounds.height / 200 : -imgBounds.height / 200;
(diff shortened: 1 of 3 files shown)


simonwiles:
The at Schlözer at `?druid=sz207mm8382` is a good one to test, but you probably know others :)

broadwell:
This does slow down both the acceleration and the apparent top speed (?) of scrolling via the arrows and <kbd>ctrl</kbd>-mousewheel panning (though the latter is still somewhat effective on my computer), but it seems necessary to facilitate the fine-grained scrolling that was requested. For what it's worth, the speeds produced do seem somewhat closer to what an actual player piano would be capable of.

I'm fine with merging this, especially considering that other options for higher-speed scrolling/seeking throughout the roll are available (e.g., the progress slider) and others are in the works (the vertical navigator).

PR #70, opened 2021-07-27
Use MIDI->control hole function map to assign labels

broadwell:
I tried to limit the `rollProfile` map to the most essential information, but its contents are still prone to changing later.

It made sense to continue to export `getNoteName()` from `utils.js` because it may be used elsewhere in the future.

FYI I have no idea what the `motor_off` and `motor_on` holes are supposed to do, but will ask.


simonwiles:
It's so nice to have this available now -- it really improves my personal experience playing with the app, and I'm certain it will for others too.

* I wonder if the `rollProfile` config should live in its own file (just for the DX -- doesn't make any difference once bundled)?
* Now that we have tooltips over on the right edge, the flow out of the viewer and get cut off :(  Upon initial inspection, OSD seems to apply `overflow: hidden` directly to two containers, which means: 1) it's hacky to wind it off again; and 2) it may be doing it for a good reason (although brute-forcing it to `visible` didn't immediately seem to blow anything up).  I'm wondering about having the label "flags" pointing to the left for the whole on the right-hand side?  Think there's probably a cheap solution here.

simonwiles:
Re. the second bullet above, now that I think properly about it, this has always potentially been an issue if zoomed in such that note holes are close to the right-hand edge of the viewer (and so the *really* cheap solution I was thinking of probably won't fly). 🤔

broadwell:
`rollProfile` now lives in its own file. It does seem cleaner this way.  
The clipping of the overlay tooltip flags does need to be addressed, I think.

src/utils.js
```diff
@@ -1,33 +1,44 @@
...
+    "F",
+    "F#",
+    "G",
+    "G#",
+  ][(midiNumber - 21) % 12];
+  return `${name}${octave}`;
+};
+
+export const getNoteLabel = (midiNumber, rollType = "welte-red") => {
+  let noteLabel = `mid_{midiNumber}`;
```
simonwiles, line 30: Typo: should be ```let noteLabel = `mid_${midiNumber}`;```

simonwiles (changes requested):
Looks good -- just the one typo.  If you could just fix that I'll merge this and we'll look at the tooltip clipping in another PR.

src/utils.js
broadwell: Yikes -- fixed now.
Btw, for a roll that exposes this functionality, I recommend `cy287wz7683` because the hole detections are so badly mangled.

simonwiles: Thanks -- I wasn't sure to what extent that was a real concern.

PR #72, opened 2021-07-29
Make note velocities available in RollViewer

broadwell:
The velocities are stored in the per-roll JSON files' `holeData` array. Given that those files all needed to be updated, it made sense to add a refresh of the MIDI files and `catalog.json` to incorporate some recent improvements in roll image parsing and metadata wrangling.

Hiding the velocity values for the 88-note rolls is a consideration, since they're always `64`. On the other hand, making them available via this fairly advanced option does at least satisfy the curiosity of any user who might be wondering what setting we used.


simonwiles:
(I've moved the `yarn.lock` commit to a new branch and will be adding a couple of extra commits before PRing it again, sorry for the confusion.)

> Hiding the velocity values for the 88-note rolls is a consideration, since they're always 64.
I think showing them is a good idea -- I wonder if we need to have them written into the `.json` for every hole, though?

Otherwise this looks great, I'm going to merge it now, and we can consider whether to crunch the `.json` files a bit later (taking the spaces out after colons and commas is worth a couple of kb even after gzip, and dropping the `"v"` keys seems to be worth about 15% on the 88-note rolls -- admittedly that's before gzip, though).

src/components/RollViewer.svelte
```diff
@@ -166,9 +166,20 @@
...
   const createMark = (hole) => {
-    const { x: offsetX, y: offsetY, w: width, h: height, m: midiKey } = hole;
+    const {
+      x: offsetX,
+      y: offsetY,
+      w: width,
+      h: height,
+      m: midiKey,
+      v: velocity,
+    } = hole;
```
simonwiles, line 176: I wonder if we should just do 
```
    const {
      x: offsetX,
      y: offsetY,
      w: width,
      h: height,
      m: midiKey,
      v: velocity = 64,
    } = hole;
```
here and then just not write the velocity values in the `.json` for non-reproducing rolls?

simonwiles (approved):

src/components/RollViewer.svelte
broadwell: Yeah, there's no reason to include the `v: 64`s in the JSON for the pianola rolls, and actually I think that for now we shouldn't expose any velocity info about them at all. Ideally, if we did so, the values would change in response to the main, treble and bass volume settings, as well as the accent button. Probably they should be similarly responsive for the reproducing rolls, but I think we have more leeway on this because you're not really supposed to touch the controls after a reproducing roll begins playing. Anyway, we can finesse all of this later.

I've pushed an update that removes the velocity values from the pianola JSON files. It should be good to go now.

PR #73, opened 2021-07-30
Dependency and linting house-keeping

simonwiles:
This PR has your commit from the other branch that includes an eslint bump that flags two a11y issues, so it also includes a fix for those.  Finally I've also removed the `eslint-plugin-jsx-a11y` plugin, since it wasn't actually operational anyway.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   root: true,
-  plugins: ["svelte3", "jsx-a11y"],
+  plugins: ["svelte3"],
   extends: ["airbnb-base", "prettier"],
   parserOptions: {
     ecmaVersion: 2020,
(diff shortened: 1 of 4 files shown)


broadwell:
This all looks good to me.

On the topic of cleanup, we should address these browser warnings at some point. No rush, though.
```
<FilteredSelect> was created without expected prop 'selectedItem' FilteredSelect.svelte.js:867:12
<RollSelector> was created without expected prop 'currentRoll' RollSelector.svelte.js:164:12
<KeyboardControls> was created without expected prop 'outside' KeyboardControls.svelte.js:514:12
```

simonwiles:
> On the topic of cleanup, we should address these browser warnings at some point. No rush, though.

Yeah, these are just dev. warnings -- they're not thrown in the build version (I think this is because they're just advisory, but it could also be because they're actually products of the dev/HMR server).  I don't think we need to do anything about them, tbh.

PR #74, opened 2021-07-30
Colorize, highlight active holes by velocity, type

broadwell:
I think the range of yellow saturations works pretty well for indicating velocity, along with some subtle scaling of the outer glow. Differentiating the control holes by color was kind of an afterthought, but could be worth keeping.


broadwell:
Closing in favor of #79 .

simonwiles:
Do you want to keep this branch?

broadwell:
We can delete it -- all of the important bits of code are also in `enhanced-hole-highlighting`, I think.

PR #75, opened 2021-07-30
Flip overlay flags to the left when near right edge of viewer

broadwell:
This keeps the overlay flags from getting clipped by the right edge of the viewer. Just allowing the flags to overflow the viewer is problematic both aesthetically and technically. The threshold is set arbitrarily at 80% of the width for now; it could be something like 50%, but I think it's slightly preferable aesthetically for the note hole flags to all point the same way at the default zoom level, and this happens when threshold = 80%.

This doesn't address the possibility that the active overlays (the labels that can appear _above_ the holes on playback) might be clipped, but that rarely if ever happens due to their positioning.

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -46,6 +46,12 @@
 
     :global(mark) {
       background-color: transparent;
+      --left-flag-position: calc(
+        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
+      );
+      --right-flag-position: calc(
+        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
+      );
 
       &:hover {
         background-color: transparent;
@@ -82,9 +88,8 @@
       content: attr(data-info);
       display: block;
       font-weight: bold;
-      left: calc(
-        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
-      );
+      left: var(--left-flag-position);
+      right: var(--right-flag-position);
       padding: 8px ($highlight-hover-outline-width + 4px) 8px 4px;
       position: absolute;
       text-shadow: 0px 0px 8px black;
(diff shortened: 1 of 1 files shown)


src/components/RollViewer.svelte
```diff
@@ -46,6 +46,12 @@
 
     :global(mark) {
       background-color: transparent;
+      --left-flag-position: calc(
+        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
+      );
+      --right-flag-position: calc(
+        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
+      );
```
simonwiles, line 54: These two variables are the same, so can we just use one?

src/components/RollViewer.svelte
```diff
@@ -82,9 +88,8 @@
       content: attr(data-info);
       display: block;
       font-weight: bold;
-      left: calc(
-        100% + #{$highlight-hover-outline-offset} + #{$highlight-hover-outline-width}
-      );
+      left: var(--left-flag-position);
+      right: var(--right-flag-position);
       padding: 8px ($highlight-hover-outline-width + 4px) 8px 4px;
```
simonwiles, line 93: The padding needs flipping too unfortunately.

src/components/RollViewer.svelte
```diff
@@ -190,6 +195,15 @@
       width,
       height,
     );
+    const imgBounds = viewport.viewportToImageRectangle(viewport.getBounds());
+    const markFractionalPosition =
+      parseFloat(offsetX + width / 2 - imgBounds.x) /
+      parseFloat(imgBounds.width);
+    if (markFractionalPosition > 0.8) {
```
simonwiles, line 202: This is a decent enough heuristic that seems safe to me and simplifies the implementation considerably, so I'm +1 on this :)

simonwiles (dismissed):
Setting both css variables and then overriding one with `auto` is an elegant solution to the problem of using JS to make dynamic changes that involve SASS variables.  Given that `--left-flag-position` and `--right-flag-position` are identical we probably don't need both.

The only difficulty is that the padding kinda needs to be flipped too.  This *could* be done with the same clever approach, most likely, but I wonder if just using a `.flag-left` class is simpler at this point?

src/components/RollViewer.svelte
broadwell: Yes! As long as it works with the mechanic of setting the "off-side" value to `auto` below, I guess. I figured there was probably a better/DRYer way to do this...

src/components/RollViewer.svelte
broadwell: Hmm yeah, considering this, someone suggested that it may be better just to set class to make these adjustments. 😄

simonwiles (approved):

broadwell:
Awesome, thanks!

PR #77, opened 2021-08-02
Persist `<RollSelector/>` facets

simonwiles:
Are they really facets if you can only select one at a time?  Probably not...


broadwell:
All good.

PR #79, opened 2021-08-04
Colorize, highlight holes according to type, settings, velocity

broadwell:
This supersedes PR #74, as well as PR #55, from which it borrows liberally for the "Highlight All Holes" feature. There are two main areas of new/updated functionality:

1) A new visualization option, "Highlight All Holes," which fully colorizes the full-viewport base SVG overlay to differentiate control holes (light green), pedal holes (orange), and note holes, which are either yellow if no velocity data is available for the roll, or mapped to a range from blue to red according to their velocities.

2) The colors described above are also applied to each hole's active highlight, along with a velocity-modulated scaling of the note holes' outer glow, subject to the following modifications:
- Velocity-based visualizations (colors and modulated glow) are only applied to the active note holes when "Display Note Velocities" and "Play Expressions" are both enabled.
- Active control holes are only highlighted when "Play Expressions" is enabled.
- Active pedal holes are only highlighted when "Use Roll Pedaling" is enabled.

The base overlay layer (visible when "Highlight All Holes" is selected) remains fully colorized even when expression and pedaling are disabled in the Roll Emulation Settings, but I think this is OK, as it reflects a consistent if rather subtle distinction between the purpose of the Visualization Settings vs. the Roll Emulation Settings. For the latter, I think it's most important that the active note/pedal/control holes are either highlighted or not when the trackerbar passes them, in accordance with the emulation settings.


simonwiles:
I've added three commits:
* to fix the "leftwards" flags when "Show Details for Active Notes" is on
* to prevent the flickering hover wars when "Show Details for Active Notes"
* to add some padding to the `<rect/>` elements for when "Highlight All Holes" is enabled

I like the look of the "Highlight All Holes" feature a lot more now -- hope you agree?

Still not thrilled about the look of the "Display Note Velocities" feature -- will what I might be able to do there next.

Also, you made a throw-away remark at one point about componentizing the hightlinging somehow (at least, I think that's what you were getting at?) -- I think this is probably a very good idea, and will have a look at doing that now.  The code for the `<RollViewer/>` component is getting unruly.  I would like to do a little refactoring of the code in this branch, and I should probably factor out the `<mark/>` elements (and perhaps event the `<rec/>` elements) into Svelte components at the same time.

simonwiles:
Also just noticed that with "Highlight All Holes" enabled you can't see the control holes being activated (before or after my additions).  Not sure this is a big problem, but wanted to mention it.

broadwell:
Great -- the "Highlight All Holes" view definitely looks a lot better now.

Factoring out the overlay elements seems like a good way to streamline the `RollViewer` code, so I'd definitely be in favor of that. Also on the topic of refactoring, we've previously discussed that some further overhaul of the `RollViewer`'s lifecycle might help to avoid the pernicious error by which the selected roll's image fails to load -- I suspect that just instructing the OSD viewer to load a new image, rather than tearing down and resetting the entire component, could alleviate this, but can't really say exactly why. But anyway, that's probably best left to another PR.

Finally, if I zoom in really close and advance the roll slowly, I can see the control holes being activated when "Highlight All Holes" is enabled, but due to the short duration of the holes and the lightness of their highlight colors, it's very difficult to see this at normal playback speeds. Probably any change that makes the colorized "Display Note Velocities" highlights more visible (if we decide to keep this feature) will also improve the visibility of the control and pedal hole highlights.

broadwell:
A couple of other oddnesses that we might as well consider now:

- Changes made to the visualization settings or the roll emulation settings do not affect the currently active/highlighted holes -- the changes only become apparent when the next set of holes is highlighted. Ideally, the colorized/yellow highlights and velocity value display for the currently active holes would toggle and/or change color when "Display Note Velocities", "Play Expressions", and "Use Roll Pedaling" are toggled on or off.

- Maybe this isn't worth addressing (it might be rather complicated), but if the roll is paused while notes are highlighted with the "Show Details for Active Notes" text appearing above them, this text remains (and at the same size) even if the roll is zoomed out and the apparent trackerbar position moves beyond the highlighted holes. The latter happens because the actual play position *hasn't* actually moved; rather it's just out of sync with the visualized position of the trackerbar on the roll, and if the roll is not manually advanced or rewound, the roll will jump back to the previous position when Play/Pause is clicked again.
Screen Shot 2021-08-04 at 9 16 36 PM

broadwell:
OK, I added one more commit that
- Removes the outer glow radius modifications for the highlights, because this effect wasn't really visible anyway.
- Ensures that the yellow active highlight color is always in effect for both the note and control/pedal holes, unless "Highlight All Holes" is NOT selected, and "Display Note Velocities" IS selected -- in this case, the colorized highlights are used. This makes the active holes much easier to see when "Highlight All Holes" is on, but doesn't really obscure any hole velocity/function info because the colorized versions of the holes are always visible when they're not active, and are still somewhat visible (around the edges, and after the animation fades) even when they are active.

I'd be happy to see other proposed enhancements to the colorized hole highlighting, but I think this is already an improvement.

Btw I know that the logic of `getHoleColor()` is becoming increasingly complicated, but on the other hand, it still works...

simonwiles:
> Factoring out the overlay elements seems like a good way to streamline the RollViewer code, so I'd definitely be in favor of that.

This might not be an option, in the end.  I don't think there's a way to create a Svelte component without immediately attaching it to the DOM, and we need detached elements to pass to OSD's `addOverlay` method.

simonwiles (approved):

simonwiles:
I'm going to approve this at this point, and ask you to merge it if you think it's ready.  This branch has got a little out of hand, so I'd like to see it go in if it there's nothing egregious, and we can address any outstanding bits in new branches.

broadwell:
OK great, I think this is good to go. I feel like some kind of commemorative/closing ceremony is called for.

 ```
    _.-"""""-._         _.-"""""-._         _.-"""""-._
   ,'           `.     ,'           `.     ,'           `.
  /               \   /               \   /               \
 |                 | |                 | |                 |
|                   |                   |                  |
|                   |                   |                  |
 |             _.-"|"|"-._         _.-"|"|"-._             |
  \          ,'   /   \   `.     ,'   /   \   `.          /
   `.       /   ,'     `.   \   /   ,'     `.   \       ,'
     `-..__|..-'         `-..|_|..-'         `-..|__..-'
          |                   |                   |
          |                   |                   |
           |                 | |                 |
            \               /   \               /
             `.           ,'     `.           ,'
               `-..___..-'         `-..___..-'
```
https://asciiart.cc/view/10770

PR #80, opened 2021-08-04
Pan Animation Handling

simonwiles:
This PR replaces the `$playingNow` store with an `$animatePan` store.  The logic is that we don't actually need to disable the panning animation while "navigating", even if that navigation happens during playback.  The approach here is to make `$animatePan` a custom store that resets itself to false after a given interval (set to approximately the animation time) *after* the last `skipToTick` call (effectively making the store behave like a singleton so that it handles the panning buttons and other elements that repeatedly fire events and update `$currentTick`).

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -63,6 +63,7 @@
     rollMetadata,
     showKeyboard,
     overlayKeyboard,
+    animatePan,
   } from "./stores";
   import { clamp } from "./utils";
   import SamplePlayer from "./components/SamplePlayer.svelte";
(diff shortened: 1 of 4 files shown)


broadwell:
This works fine for me. The self-resetting store is quite nifty, and I'm relieved to see that the animation time calculations really do operate in the realm of basic arithmetic.

One lingering consequence of the immediate `panTo()` calls is that small corrective jumps after the main animation are noticeable when dragging the viewer or using the up/down buttons during playback. But this mode of interaction isn't really desirable and probably not something we should go out of our way to accommodate -- which is another way of saying that I'm fine with the observed behavior.

PR #81, opened 2021-08-04
Zoom button handling

simonwiles:
This PR tidies up the handling of the pan buttons when the button is held down, and applies the identical solution to the zoom buttons.


broadwell (approved):
Great, the zoom in/out button behavior is much appreciated.

broadwell:
Thoughts about moving the up/down buttons, and if so, to where? I think that anyway they don't overlap with the play point/trackerbar plane would be preferable. Maybe just 3/4 of the way up the right (or left) side?

PR #82, opened 2021-08-06
Enable use of pedaling, treble and bass accent data for 88-note rolls

broadwell:
It turns out that 88-note rolls can contain both treble- and bass-specific accents, as well as some very rudimentary pedaling indications, so we figured that we might as well make these available in the app. "Play Expression" and "Use Roll Pedaling" are now available for 88-note rolls, but not enabled by default.

The roll MIDI files now contain pedaling and velocity indications, the latter of which are propagated to the JSON files. There are really only two possible velocity values (75 and 95, for now), so encoding them for each note hole in the JSON isn't terribly efficient. This would almost certainly be the easiest kind of note-MIDI-based "expression emulation" to do directly in the app...


src/App.svelte
```diff
@@ -194,6 +197,8 @@
         $rollMetadata = { ...$rollMetadata, ...metadataJson };
         if (metadataJson.holeData)
           buildHolesIntervalTree(metadataJson.holeData);
+        $playExpressionsOnOff = $isReproducingRoll;
+        $rollPedalingOnOff = $isReproducingRoll;
```
simonwiles, line 201: I suppose this will be a full-blown UI configuration function at some point, but this is the correct place to do it and this is perfect for now.

src/stores.js
```diff
@@ -29,7 +29,12 @@ const createSetStore = () => {
...
 export const rollHasExpressions = derived(
   rollMetadata,
-  ($rollMetadata) => $rollMetadata.ROLL_TYPE === "welte-red",
+  ($rollMetadata) => $rollMetadata.ROLL_TYPE !== "65-note",
+);
+export const isReproducingRoll = derived(rollMetadata, ($rollMetadata) =>
+  ["welte-red", "welte-green", "welte-licensee"].includes(
+    $rollMetadata.ROLL_TYPE,
+  ),
 );
```
simonwiles, line 38: Nice :)

simonwiles (approved):
Very happy with all this.  The velocity values in the JSON for the 88-note rolls is fine, tbh -- with only two values it compresses extremely well, and doing otherwise would be a fudge.  Now if we're computing expressions in the client, that's a different matter, of course :)

PR #84, opened 2021-08-11
Miscellaneous Updates

simonwiles:
* Move the vertical pan buttons to the top
* Set the max value for the tempo slider to 300%
* Fix the position of the nav strip label for upwards-scrolling rolls
* Prevent the tracker bar indicator from extending over the nav strip
* Use ODS's built-in `scrollToZoom` (disables scroll-to-zoom behaviour on the nav strip, preventing odd interaction effects when zooming and panning at the same time)

+ some linting, housekeeping, and code cleanup.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -63,7 +63,6 @@
     rollMetadata,
     showKeyboard,
     overlayKeyboard,
-    animatePan,
     isReproducingRoll,
     playExpressionsOnOff,
     rollPedalingOnOff,
@@ -133,7 +132,6 @@
 
   const skipToTick = (tick) => {
     if (tick < 0) pausePlayback();
-    $animatePan = true;
     $currentTick = tick;
     updatePlayer(() => midiSamplePlayer.skipToTick($currentTick));
   };
--- a/src/components/BasicSettings.svelte
+++ b/src/components/BasicSettings.svelte
@@ -79,7 +79,7 @@
     <span>{($tempoCoefficient * 100).toFixed(0)}%</span>
     <RangeSlider
       min="0.1"
-      max="4"
+      max="3"
       step=".001"
       bind:value={$tempoCoefficient}
       name="tempo"
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -295,7 +295,7 @@
     osdNavDisplayRegion.dataset.label = ($playbackProgress * 100).toFixed(1);
     osdNavDisplayRegion.classList.toggle(
       "label-above",
-      $playbackProgress > 0.5,
+      scrollDownwards ? $playbackProgress > 0.5 : $playbackProgress < 0.5,
     );
   };
 
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -314,6 +314,32 @@
     });
   };
 
+  const updatePosition = ({ target, immediately }) => {
+    if (immediately) return;
+
+    const { centerSpringY } = viewport;
+    centerSpringY.animationTime = 1.2;
+
+    clearInterval(animationEaseInterval);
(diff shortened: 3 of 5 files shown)


broadwell:
All of the other changes check out fine for me, but I figured I should note that there's still some odd behavior around zooming and panning with the navigator.
- First of all, it's not longer possible to zoom by using the scroll wheel while hovering over the navigator, which I think is an improvement.
- However, it's still possible to get some extreme panning behavior by zooming out in the main viewer and then dragging the navigator in one direction or the other before the zoom animation has completed.
ExtremePanning
Setting `constrainDuringPan: false` seems to squash the worst of this behavior, but it has some other effects that are less desirable (e.g., extreme slippiness), so it would be nice to find another way to suppress the zoom+pan amplification.

I'm OK with merging this branch as is, but note the above as a lingering quibble that we hopefully can address later.

simonwiles:
More updates:

* `$animatePan` has been dropped in favour of manually adjusting the animation time of the vertical animation spring.
* use of OSD's native `scrollToZoom` functionality has been dropped (again) -- it was causing a couple of problems, including causing unwanted horizontal panning (debatable, perhaps, but I personally didn't like it at all), but most importantly it was causing the viewer and app position to get out of sync on zoom.  This could have been hacked, of course, but it seems cleaner in this case to just handle the zooming-on-scroll ourselves, especially since we were doing that originally anyway.
* OSD's navigator on-click behaviour has been overridden with a bespoke implementation that is now shared with the `pan` handler.

* A lot more reorganizing and refactoring

Still more to come, I think.

broadwell:
The code looks much more consistent and streamlined now. Regarding functionality, I'm OK with the limited horizontal panning. I also kind of preferred when the navigator didn't respond to mouse zoom/deltaY events at all (so it really is just a fancy scrollbar in that case) but I'm not sure if that can be implemented easily in the current regime.
Actually I think a lot of my discontent with the zooming behavior (including the zoom+pan "freakouts") are partly due to the high mousewheel/2-finger zoom sensitivity on MacOS browsers -- it seems that it is more subdued on other platforms, and also that it's difficult to normalize:
https://twitter.com/openseadragon/status/1161713525355454464
https://github.com/openseadragon/openseadragon/issues/1820
Anyway, I'm still a bit bothered by the how the note highlighting races ahead of the trackerbar during rapid panning, but I can live with it as long as it doesn't happen during playback or when using the up/down buttons.

simonwiles:
Thanks for the update -- I reckon I've just got a couple more commits on this branch before it's ready for merging (spent the morning wrestling with canvas dragging).

> Anyway, I'm still a bit bothered by the how the note highlighting races ahead of the trackerbar during rapid panning, but I can live with it as long as it doesn't happen during playback or when using the up/down buttons.

I'm not sure if you were testing using the most recent commits?  Basically all of the interactions are explicitly pass either `animate=true` (significant movements will cause the image panning to take a while to catch up with the player position and the highlighting) or `animate=false` (everything will happen in sync, but no smooth animation or easing). 

At present, it's configured like this:
* canvas point-and-click dragging / flicking: `animate=true`
* navigator click: `animate=true`
* panning buttons / mousewheel scrolling: `animate=false`
* navigator / progress bar dragging: `animate=false`

These are the settings I've arrived at -- but they can be changed easily (I hesitated to make the user-configurable, but I suppose that's easy enough if we consider the trade-off vs. settings overload is favourable).

broadwell:
I'm OK with the other settings, but still do kind of prefer `animate=true` for canvas point-and-click dragging / flicking. The feeling is definitely less slick than to the animated version, but the experience overall seems more similar to what it would be like to actually advance a roll manually. But for now I think it's enough that the ctrl+mousewheel panning gives an indication of what the non-animated dragging would be like -- we can just ask the PIs which they prefer.

PR #86, opened 2021-08-16
Legend / info panel / ?

simonwiles:


simonwiles:
This is a draft PR for now -- I'm not sure what to do with this next, and so I want some extra eyes on it.

I've been thinking about things like elapsed/remaining/total time (based on current tempo), same for physical extent (in ft. and in., or whatever), clickable unit conversion (for the civilized metric-using world) etc. etc.

I also don't know that the striking through of "Pedal Holes" and "Control Holes" when they're disabled is great.

Comments and ideas very welcome :)

broadwell:
I think the functionality is great, especially the ability to slide it out diagonally. All of my feedback so far just consists of quibbles about the content:

- Maybe just have the Tempo readout appear if the right-hand panel is closed?
- "Pedal Holes" could be just "Pedals", "Note Holes" -> "Notes", "Control Holes" -> "Controls"
- The velocity color legend could be perhaps just a horizontal map, with maybe 0% and 100% on either end -- note sure how much value there is in giving the percentage mapping of each shade.
- I'm kind of on the fence about the strikethroughs; just removing the color from the boxes may be sufficient.

simonwiles:
> All of my feedback so far just consists of quibbles about the content
Thanks, that's exactly what I was after :)  And I agree enthusiastically with all your points (except maybe for the one about remove the word "Holes" -- that one I am willing to agree with, but not enthusiastically :)).  I think I was planning to remove tempo as a percentage altogether, and I was also considering removing the pedals/controls lines altogether if they weren't active.  Really not too sure.  The horizontal map seems like a clear win -- I'll try to put something together.

broadwell:
The code from this PR has now been integrated into the `develop` branch, albeit in somewhat atomized form -- see #190 .

PR #87, opened 2021-08-19
Enable configuration of piano sample volumes and reverb

broadwell:
This is ready for feedback if not review and merging. Except for the number of samples, all of the settings can now be changed on the fly via the "piano" menu icon. Changing the number of samples doesn't work at all yet; as far as I can tell, when using the stock ToneJS piano module it always requires recycling the entire piano component in order to (re)-load the samples, which might require some extra finessing within the Svelte environment.

I'll continue to experiment with other reverb/environmental acoustics packages and modules, but the standard ToneJS Reverb module might have the best achievable balance between performance/overhead and sound quality.


broadwell:
Update: changing the number of velocity samples on the fly now works, though it's a bit clunky and could be better integrated into the app (e.g., probably a notification should be displayed while the samples are reloading, especially if it happens during playback). Supporting the full range of velocities, in addition to potentially taxing the user's memory if a large number is chosen, also complicates deployment a bit since it involves shipping around ~100 MB of MP3s. But I think it's extremely likely that the only way to assuage concerns about the playback sound quality of the app is by providing at least the option of using a high number of sample velocities + reverb.

src/components/SamplePlayer.svelte
```diff
@@ -73,7 +83,6 @@
       softOnOff.set(pedals.includes(SOFT_PEDAL));
     } else {
       sustainOnOff.set(false);
-      piano.pedalUp();
```
simonwiles, line 76: What's going on here?  Is this connected to the issue that's addressed in the `no-sustain-on-rewind` branch?  Did it creep into the "Update sample volumes on change" commit by accident?

src/components/SamplePlayer.svelte
```diff
@@ -34,19 +40,23 @@
...
 
-  const piano = new Piano({
+  piano = new Piano({
     url: "samples/",
-    velocities: 4,
+    velocities: $sampleVelocities,
     release: true,
     pedal: true,
-    maxPolyphony: 64,
+    maxPolyphony: 256,
```
simonwiles, line 48: Why not indeed? :)  In all seriousness, though, is there anything to be concerned about here?  Why does the ability to set a `maxPolyphony` even exist, and why not default it to a really high value?  Do we need to do some testing (somehow?) to make sure this is "safe"?

simonwiles (commented):
It's great to see this basically working now.  I have a couple of things for discussion:

1. Having an Audio Settings panel works great, but having three panels is now exposing the limitations of the quick-n-dirty job I did on the panel switcher (the animation when switching between panels 2 and 3 is broken, the code could use some DRYing up and refactoring, too).  I was starting to wonder about redoing that lately anyway.  I don't think it's a blocker for this PR, but there's plenty that could/should be done here (including rethinking the organization of what we already have).  I was thinking about tabs or an accordion or something.
2. The code in `SamplePlayer.js` looks like it's crying out for us to subclass the `Piano` class.  I think it makes sense to keep our `tone-piano.js` as a drop-in replacement for `@tonejs/piano` as it is now (I made one small performance optimization, that's all), such that the npm module could be subbed back in at any time (SSR issues notwithstanding).  However, subclassing it would allow all this functionality to be grouped together in a more maintainable way, and would allow address the issue of pedal events that occur with the same timestamp in a nice clean way too.  I'd be happy to take a look at that, if it's okay with you, but it could be before or after this PR is merged.
3. Finally there is a lot going on here in terms of changes to the playback.  It all sounds very different to me by default now.  Increasing the reverb seems to decrease the volume in some cases?  In any event, seems like there's lots to be discussed.  There's also the issue of appropriate warnings (the page uses > 3gb of RAM on my system with 16 velocities!) -- even if we're not ready to add "explanations" yet -- and so on.  All of which is to say I think we both a) need to get this in front of the PIs; and b) might be better not to push it right into production.  So is this the time for a staging branch which is automatically deployed to Netlify or similar?

src/components/SamplePlayer.svelte
broadwell: Yeah this is spillover from that other branch. My understanding is that the line is entirely redundant because setting `sustainOnOff` to `false` triggers a call to `piano.pedalUp()` anyway.

src/components/SamplePlayer.svelte
broadwell: This has been a known shortcoming for a while and I figured it should be addressed during any kind of overhaul of the piano module/functionality. Examples with the current roll set are quite rare; actually the only one I've found is at the end of https://sul-cidr.github.io/pianolatron/?druid=yx536mq2915 -- if you start playback around 92%, it runs out of polyphony allotment around 97%. But it could conceivably happen in any instance (e.g., during a pianola performance) when heavy sustain pedaling causes more than 64 notes to be held at the same time.

The limit seems to derive from this code:
https://github.com/Tonejs/Tone.js/blob/053b5d4397b595ea804b5d6baf6108158c8e0696/Tone/instrument/PolySynth.ts#L168

So we can't get rid of the limit entirely without modifying the ToneJS codebase, but on the other hand, it's just a logical limit; if we set it to a large number, it doesn't actually result in a massive amount of memory being preallocated. So maybe we could set it to something like `Number.MAX_SAFE_INTEGER`?

simonwiles: Thanks for hunting down that reference.  I can't see anything to be concerned about there, but I still wonder what's even the point of having the `maxPolyphony` argument at all, and it unnerves me ever-so-slightly that there might be something I'm not seeing.  I'm happy with just setting it beyond reasonable bounds (perhaps just `Infinity` is simplest) and keeping our eyes open for issues, I suppose.

broadwell:
I agree with all of your points. As an initial, blanket response, I'll say that I do think this branch would be a perfect candidate for a staging/dev version of the app to be deployed on Netlify. The UI and piano component definitely could use further refinement, but particularly with respect to point 3, I think it would be worthwhile to get a version in front of the PIs ASAP so that they can explore and provide feedback on the audio parameters the branch exposes.

src/components/SamplePlayer.svelte
simonwiles: Yeah, that seems right.  Technically setting `sustainOnOff` to `false` will only result in a call to `piano.pedalUp()` if `sustainOnOff` was previously `true` -- but I can't think of a reason why that distinction would matter right now.  I guess it was probably just an error on my part when I put it in in the first place.

simonwiles:
Merged to `staging` and deployed to https://pianolatron-staging.netlify.app/

I've disabled deploy previews going forward -- only explicit pushes/merges to `staging` will be deployed.

src/components/SamplePlayer.svelte
broadwell: Yes it's rather puzzling. It's almost as if the ToneJS PolySynth were written to run on a memory-constrained embedded controller... but that seems like an odd fit for a Javascript (or TypeScript) package. Anyway I've added a commit that sets the polyphony limit to `Infinity`, although the version on `staging` that uses `256` should be fine for now.

simonwiles: Perhaps a hold-over from a different way of doing things previously :shrug:  Anyway, happy with this for now, at least.

simonwiles:
Right, this has seen a lot of updates and is ready to come back to you for another review.

Since your initial PR, the major changes are:

1. The creation of `pianolatron-piano` which subclasses `tonejs-piano` and exposes the additional functionality we need (you've seen most of this, I think).
2. Changing the number of sample velocities can be done without reloading the `Piano()` instance, and the loading spinner is displayed while the playback is paused as the samples are loaded.
3. A warning is presented when attempting to increase the number of velocities over the default value of 4.
4. The piano settings (or is that the audio settings?) are persisted in `localStorage`.

To achieve these changes a fair bit of other yak shaving had to be done.  Some of the interesting bits:
* The loading spinner is now it's own component, and has been restyled a bit (and now fades out, etc.).  Of particular note is that the completely statically rendered `index.astro` now uses the svelte component directly (but statically renders it, unlike the `<Pianolatron/>` component which is hydrated `client:only`.  This is quite cool, imo.
* The `<Notification/>` component has a more useful default style, and the `notify()` call can now accept an array of `actions` which are rendered as buttons with associated callbacks.

One thing that might be regarded as outstanding is the lack of a "don't show this again" checkbox for the sample velocities warning dialogue.  I couldn't decide how important this was -- it will need more retooling of the `<Notification/>` component than I'm willing to do tonight, and I wanted to get this PR raised.  I'll let you decided where it's a blocker for this PR being merged :)

src/components/SamplePlayer.svelte
```diff
@@ -258,6 +312,9 @@
   $: $tempoCoefficient, updatePlayer();
   $: $useMidiTempoEventsOnOff, updatePlayer();
   $: $rollPedalingOnOff, updatePlayer();
+  $: piano.updateVolumes($pianoSettings.sampleVolumes);
```
broadwell, line 315: These hierarchical reactive variables are causing trouble here, it seems. A change to any of `$pianoSettings`'s children triggers **both** `piano.updateVolumes()` and `piano.updateReverb()` to run, more or less at the same time, regardless of which child actually has been updated. This causes the playback and scrolling to become pretty choppy when one of the sample volume sliders is dragged while a roll is playing (on my machine, at least), and also causes progressively more static/distortion to be heard in the playback -- compare this to the peaceful scene when a sample volume slider is scrolled during playback on [staging](https://pianolatron-staging.netlify.app).

Obviously both updates shouldn't be happening at the same time, and one way to avoid it would be to deconstruct `$pianoSettings` or maybe change how the reactive triggers for its children are defined. The underlying mechanism that seems to be causing harm to the app, however, is apparently the rapid-fire disconnection/destruction, creation, and connection of a Reverb module, which on my computer at least can cause the audio to become progressively more distorted until the browser itself needs to be closed. Ensuring these events only happen when the Reverb slider is moved, and only updating the slider value when it has moved 5%, seems sufficient to avoid the worst-case scenario, but I'd feel better if there was a more foolproof way of handling it.

src/components/SamplePlayer.svelte
```diff
@@ -85,13 +89,63 @@
...
+      return loadingSamples;
+    });
+  };
+
+  const updateSampleVelocities = () => {
+    if (
+      $pianoSettings.sampleVelocities > 4 &&
+      $pianoSettings.sampleVelocities > piano.loadedVelocities
+    ) {
+      notify({
```
broadwell, line 126: Could/should we disallow changes to the Sample Count slider while this dialog is open? It's a little weird to be able to move the slider -- and have it actively reload the samples in the background when the slider goes below 5 -- while we're supposed to be holding up and thinking long and hard about what we're doing. But the current behavior isn't really harmful, so I'd also be OK with keeping it as is.

broadwell (commented):
This is great, excepting the two issues noted in the code -- one much more consequential than the other.

src/components/SamplePlayer.svelte
simonwiles: Yeah, I'd not noticed that, and I'm not sure it's really okay, tbh.  I've updated the `<Notification/>` component to accept a `modal` property, and applied it to the sample velocities warning -- I think this should have the desired effect.  Turned out to be simpler than I'd imagined it might prove to be, anyway.

src/components/SamplePlayer.svelte
simonwiles: Yeah, both of these need addressing, thanks (though for the record I'll note that there is no user-discernible difference between this branch and staging on my set up here, so I'm grateful you caught it).  Working on it now, but have meetings through the morning, so if it's awkward it might be this afternoon now.

simonwiles: Okay, I think these issues should be fixed now.  I did deconstruct the `$pianoSettings` store (although this necessitated a bit of work to the persistence affordances), and to address the issue of the reverb being disconnected and reconnected to frequently I set the slider only to "bite" when the slider is released (as with the "Sample Count" slider).  There may be a better option here -- perhaps just some visual feedback to indicate that it's *not* being updated completely "live" would be adequate and/or necessary?

In the process I discovered and fixed a couple of other small bugs: the sample volumes weren't reset properly when the "reset all settings" button was used; and the mousewheel handling was broken for the "Sample Count" slider.  I took the approach of just disabling it in the latter case.

broadwell: Excellent, the distortion problems on Mac OS + Firefox and Chrome (Firefox has persistently been a bit worse than Chrome, btw) are gone now. I believe they were also present on Windows, at least in some situations. Anyway I think that making the reverb slider less "bitey" should be sufficient to avoid the runaway distortion syndrome, given that what was really inflaming the situation was that the much more responsive keybed sliders were inadvertently yoked to the reverb disconnect/reconnect.

The new modal behavior seems right to me, now, and I don't miss the mousewheel bindings for the Sample Count slider. I'm not really sure there's a need for *any* of the sliders to accommodate mousewheel events. It seems a bit inconsistent that currently the Piano Sample Volumes sliders respond to mousewheel events while the Volume, Tempo* and Progress sliders in the default settings panel do not. If there's agreement that the mousewheel functionality isn't really useful for any of these sliders, them maybe it should be disabled for all of them, for consistency's sake?

simonwiles: > The new modal behavior seems right to me, now, and I don't miss the mousewheel bindings for the Sample Count slider. I'm not really sure there's a need for any of the sliders to accommodate mousewheel events. It seems a bit inconsistent that currently the Piano Sample Volumes sliders respond to mousewheel events while the Volume, Tempo* and Progress sliders in the default settings panel do not. If there's agreement that the mousewheel functionality isn't really useful for any of these sliders, them maybe it should be disabled for all of them, for consistency's sake?

Well with the exception of the "Progress" slider, and now the "Sample Count" and "Reverb" sliders, they're all supposed to have the same mousewheel-enabled behaviour... but it's always been a bit flaky, and I have an update on a forthcoming branch that makes it more stable and just nicer all-round.  We can pick this discussion back up when I PR that branch (soon, I hope).

PR #88, opened 2021-08-20
Astro

simonwiles:
Migrate to the Astro framework for SSR purposes.

If all has gone well, this should be entirely transparent to both local and `gh-pages` use; i.e., `yarn dev` should result in a development server at `https://localhost:8080/` and `yarn gh-deploy` should result in a functionally-identical deployment to https://sul-cidr.github.io/pianolatron/.


simonwiles:
This took a little longer than expected today, for reasons mostly unrelated to the PR itself.  Ideally I'd like to get this in before any new feature branches (including #87 if that's okay).  To test properly I did a full deploy to https://sul-cidr.github.io/pianolatron/, and I've actually just left it up there (I hope that wasn't a poor decision...), so you can go ahead and use it there.  It should be exactly the same from a users p.o.v. -- just bundled and browser-prefixed etc. and shipping considerably less javascript.

src/catalog.json
```diff
@@ -3254,4 +3254,4 @@
     "title": "Youmans/Sommer - Hallelujah!: fox-trot from \"hit the deck\"",
     "type": "welte-red"
   }
-]
\ No newline at end of file
+]
```
broadwell, line 3257: I'll update the metadata script to add a newline here.

broadwell:
Wow, this must have been a lot of work! I'm looking forward to adding some "content" pages.

Everything looks fine on [sul-cidr.github.io/pianolatron](sul-cidr.github.io/pianolatron), and also when I run it locally at localhost:8080 via astro. Building and serving the static site from the Python `http.server` module and [on Netlify](https://angry-wiles-a92181.netlify.app/) also work fine. The only quibble with the latter two is that it tries to look for the CIDR logo at localhost:8080, so that little bit of the footer is unhappy.

I haven't had any luck viewing the statically built version from servers running Apache -- they trip up while trying to load `_astro_frontend/hydrate/only.js` -- but I don't think this is worth holding up the merge; it may just be a local config issue, and in any case, Apache isn't in our critical path at all.

broadwell:
Huh, trying to load the CIDR logo from `http://localhost:8080` also prompts a concerned email from Netlify about possible "Insecure mixed content" on the site. Thanks, Netlify...

src/catalog.json
simonwiles: Thanks -- `yarn lint` will complain otherwise (the salient point here is that I moved `catalog.json` into the `src/` tree and so now it's subject to the linter :)).

simonwiles:
> Everything looks fine on sul-cidr.github.io/pianolatron, and also when I run it locally at localhost:8080 via astro. Building and serving the static site from the Python http.server module and on Netlify also work fine. The only quibble with the latter two is that it tries to look for the CIDR logo at localhost:8080, so that little bit of the footer is unhappy.

> I haven't had any luck viewing the statically built version from servers running Apache -- they trip up while trying to load _astro_frontend/hydrate/only.js -- but I don't think this is worth holding up the merge; it may just be a local config issue, and in any case, Apache isn't in our critical path at all.

> Huh, trying to load the CIDR logo from http://localhost:8080 also prompts a concerned email from Netlify about possible "Insecure mixed content" on the site. Thanks, Netlify...

All these issues are related to the same thing -- Astro's `buildOptions.site` in `astro.config.mjs`.  If you set this (or build with `yarn build --site http(s)://mysite.com`) then this should all work nicely.  The only caveat is that it's broken with subfolder deployments.  My work around for this until it's fixed in Astro (I've opened an issue) is to post-process the built `index.html` with `sed`... (see https://github.com/sul-cidr/pianolatron/blob/main/bin/gh-deploy.sh#L23-L24).  For a small site (if not necessarily a small app) like ours, this is acceptable -- as we add more pages they may need the same treatment -- although part of the point of Astro is that if we add static pages that don't require JS they won't ship any at all and this will be moot.

I've not set `buildOptions.site` permanently in `astro.config.mjs` in the repo for a couple of reasons.  The main one is that I used a bit of a kludge at https://github.com/sul-cidr/pianolatron/blob/main/src/layouts/footer.astro#L6 to get the image to work in a subfolder deployment, and this is why if you run `yarn build` without a value set in `astro.config.mjs` or via `--site` you get http://localhost:8080 hardcoded into the HTML (I think this needs to be addressed better in Astro -- I opened another issue for this).  The other reason is that it leaves a bit more flexibility -- it's not altogether clear to me that the gh-pages address *should* be written into the build options, tbh -- having it at https://github.com/sul-cidr/pianolatron/blob/main/package.json#L11 is arguably a little cleaner.

simonwiles:
>Wow, this must have been a lot of work!

Oh, and yeah, if you include all the time spent working on a SvelteKit version and then trying Astro and then working with the rebuild of `@tonejs/piano` and the research part.  But, as you can see, in the end there's not all that much to see and it works very well.

The main thing is the ability to use `client:only` on the main `<Pianolatron/>` component for the simplest hydration implementation -- I have a repo where I used an earlier version of Astro (and a similar SvelteKit one) which has dozens of commits where I was refactoring to access browser APIs as lazily as possible in the node SSR phase.  I was expecting to have to backport them all to this repo (which would have involved resolving copious amounts of merge conflicts, for sure), but almost none of it was required in the end! :tada:

PR #90, opened 2021-08-23
Custom Keybindings

simonwiles:
UI for configuring custom keybindings.  Allows resetting to defaults, and has some simple checks for conflicts and inappropriate keys.

### DIFF
--- a/src/components/KeyboardShortcuts.svelte
+++ b/src/components/KeyboardShortcuts.svelte
@@ -354,19 +354,19 @@
     REWIND: stopApp,
     FORWARD: () =>
       keydownRepeatAction(() =>
-        updateTickByViewportIncrement(/* up = */ $scrollDownwards),
+        updateTickByViewportIncrement(/* up = */ !$scrollDownwards),
       ),
     BACKWARD: () =>
       keydownRepeatAction(() =>
-        updateTickByViewportIncrement(/* up = */ !$scrollDownwards),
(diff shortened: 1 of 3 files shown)


simonwiles:
This should probably wait on a `LocalStorage`-backed persistence layer (and updating the defaults and adding new commands will come first, too), but I'd like if if you could take a look at this when you're able.  There are probably still some sharp edges.

broadwell:
This looks great to me so far! The only issue I've noticed is that the display names of the keys don't fit in the buttons in the dialog or in the UI proper if they involve multiple characters.
Screen Shot 2021-08-24 at 2 08 46 PM

simonwiles:
> This looks great to me so far! The only issue I've noticed is that the display names of the keys don't fit in the buttons in the dialog or in the UI proper if they involve multiple characters.
> Screen Shot 2021-08-24 at 2 08 46 PM

Thanks.  This is addressed (to some extent) in #91 which has just gone in.  There will likely be edge cases where the text creates a "button" so long as to cause problems, but in this particular case should <kbd>CapsLock</kbd> just be a banned key?  On the other hand, why not allow it?  (I actually have <kbd>CapsLock</kbd> mapped to a fourth modifier key for extra keyboard shortcuts on my plasma desktop, partly because I find it's traditional function utterly useless and an occasional hindrance when I hit it by accident!)

PR #91, opened 2021-08-23
Keyboard Commands Update

simonwiles:
This PR updates the keyboard command keystrokes to Kumaran's suggested defaults, and creates new keyboard commands where appropriate.  The only bit that's not done is the
```
Tempo increment bigger: W
Tempo increment smaller: Q

Volume increment bigger: [
Volume increment smaller: ]
```
bit, because I'm not 100% sure what to do with that (including -- but not limited to -- from a UI perspective).  Three levels, as now, but with some kind of a toggle switch for each slider?  Does the volume increment apply to bass and treble adjustments too?

Another rough edge:  I'm not entirely sure how/whether to deal with multiple keystrokes mapping to the same commands.  This comes up in two places in the immediate instance:
* the `keyCode`s for numbers are different when the events originate from a numberpad as opposed to the top row of a QWERTY keyboard -- Kumaran's requests make it fairly clear he has the top row in mind, so that's what's assigned here, but <kbd>6</kbd>,<kbd>7</kbd>,<kbd>8</kbd>,<kbd>0</kbd> etc. on the numpad don't work the same way;
* <kbd>6</kbd> and <kbd>8</kbd> pan forward and back in a way that it seems appropriate that the up and down arrows should too.

Mapping multiple keys to the same output is straightforward enough at this point, but gets a lot more complicated with the configurator-ifier thing.

Anyway, thoughts welcome.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -300,7 +300,7 @@
   {/if}
 </div>
 <SamplePlayer bind:this={samplePlayer} />
-<KeyboardShortcuts />
+<KeyboardShortcuts {playPauseApp} {stopApp} />
 <Notification />
 
 <svelte:window
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -99,6 +99,9 @@
   let startPlayback;
   let resetPlayback;
 
+  let rollViewer;
+  let updateTickByViewportIncrement;
+
   const rollListItems = catalog.map((item) => {
     const [number, label] = item.label.split(" ");
     return {
@@ -251,6 +254,7 @@
   $: playbackProgress.update(() =>
     clamp($currentTick / midiSamplePlayer?.totalTicks, 0, 1),
   );
+  $: if (rollViewer) ({ updateTickByViewportIncrement } = rollViewer);
 </script>
 
 <div id="app">
@@ -270,6 +274,7 @@
     {#if appReady}
       <div id="roll">
         <RollViewer
+          bind:this={rollViewer}
           imageUrl={currentRoll.image_url}
           {holesByTickInterval}
           {skipToTick}
@@ -300,7 +305,7 @@
   {/if}
 </div>
 <SamplePlayer bind:this={samplePlayer} />
-<KeyboardShortcuts {playPauseApp} {stopApp} />
+<KeyboardShortcuts {playPauseApp} {stopApp} {updateTickByViewportIncrement} />
 <Notification />
 
 <svelte:window
--- a/src/components/BasicSettings.svelte
+++ b/src/components/BasicSettings.svelte
@@ -49,7 +49,11 @@
     />
   </div>
   <div class="control">
-    <span>Bass Volume:</span>
+    <span
+      >Bass Volume:
+      <kbd class:depressed={$activeShortcutKeys.bassVolumeDown}>e</kbd>↓
+      <kbd class:depressed={$activeShortcutKeys.bassVolumeUp}>4</kbd>↑</span
+    >
     <span>{$bassVolumeCoefficient}</span>
     <RangeSlider
       min="0"
@@ -60,7 +64,11 @@
     />
   </div>
   <div class="control">
-    <span>Treble Volume:</span>
+    <span
+      >Treble Volume:
+      <kbd class:depressed={$activeShortcutKeys.trebleVolumeDown}>p</kbd>↓
+      <kbd class:depressed={$activeShortcutKeys.trebleVolumeUp}>0</kbd>↑
+    </span>
     <span>{$trebleVolumeCoefficient}</span>
     <RangeSlider
       min="0"
--- a/src/components/BasicSettings.svelte
+++ b/src/components/BasicSettings.svelte
@@ -36,8 +36,8 @@
   <div class="control">
     <span
       >Volume:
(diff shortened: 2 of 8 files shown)


broadwell:
All of the new changes work for me, and I'd be fine with merging this as is for now, though we'll definitely need to round-trip it past the PIs, especially considering the uncertainty around some of the new specs.

I'll ask KA if he intends for the <kbd>w</kbd>/<kbd>q</kbd> and <kbd>[</kbd>/<kbd>]</kbd> commands to function in the same way as the <kbd>shift</kbd> and <kbd>ctrl</kbd> modifiers, and if they're meant to replace or supplement that functionality (and point out that there are significant UI considerations involved). There's also the added wrinkle that previously he had hoped that <kbd>left</kbd> <kbd>shift</kbd> and <kbd>ctrl</kbd> would apply only to tempo changes, whereas the <kbd>right</kbd> modifiers would apply only to volume changes -- but I don't see how this would work with the new mappings, which place volume and tempo changes in both hands. It also would potentially make the custom key bindings even more complicated.

I think it's intended and OK for the number key mappings to apply only to the numbers on the QWERTY keyboard, rather than the numpad.

I also wonder and will ask whether it's preferred for the forward/backward keys <kbd>6</kbd>/<kbd>8</kbd> to invoke the fine-grained scrolling of the up/down buttons, or the longer-distance skipping of the arrow keys (that's how they work on my system, at least). I had thought that something more akin to a "skip forward/backwards 5 seconds" kind of functionality was desired, but now I'm really not sure.

Finally I suspect (but should confirm) that the desired rewind button is actually the <kbd>backspace</kbd> aka "backwards delete" key (the one on the main QWERTY block of English keyboards, not the function block to the right) -- note that they're both oh so helpfully labeled <kbd>delete</kbd> on Mac keyboards.

broadwell:
> Finally I suspect (but should confirm) that the desired rewind button is actually the backspace aka "backwards delete" key (the one on the main QWERTY block of English keyboards, not the function block to the right) -- note that they're both oh so helpfully labeled delete on Mac keyboards.
Following up on this: (most) Mac laptop keyboards have no proper "delete" (aka "forward delete") key at all, so I strongly suspect that the PIs intend for the rewind button to be mapped to backspace/backwards delete.

simonwiles:
>I'll ask KA if he intends for the w/q and [/] commands to function in the same way as the shift and ctrl modifiers, and if they're meant to replace or supplement that functionality (and point out that there are significant UI considerations involved). There's also the added wrinkle that previously he had hoped that left shift and ctrl would apply only to tempo changes, whereas the right modifiers would apply only to volume changes -- but I don't see how this would work with the new mappings, which place volume and tempo changes in both hands. It also would potentially make the custom key bindings even more complicated.

Yeah, I think we need some clarification here.  If there's a strong case for a clear specification, then I'm fine with doing whatever it takes to support it.

> I also wonder and will ask whether it's preferred for the forward/backward keys 6/8 to invoke the fine-grained scrolling of the up/down buttons, or the longer-distance skipping of the arrow keys (that's how they work on my system, at least). I had thought that something more akin to a "skip forward/backwards 5 seconds" kind of functionality was desired, but now I'm really not sure.

I'd thought about that a little.  Kumaran's use of "streaming/incrementally" as a qualifier to these functions led me to the conclusion that he had something like the up and down panning buttons in mind.  (I'd also considered whether we might try some kind of facility to leave a mark and rewind to that mark on a keystroke, or even to loop a section -- it's all quite doable, of course, but it's definitely feature creep.)

>Finally I suspect (but should confirm) that the desired rewind button is actually the backspace aka "backwards delete" key (the one on the main QWERTY block of English keyboards, not the function block to the right) -- note that they're both oh so helpfully labeled delete on Mac keyboards.

I think this is probably clear enough that we should just go ahead and do it.  The Mac "Delete" key fires a `Backspace` event, do I have that right?

broadwell:
Re: the forward/backward keys -- I'll just ask if the current behavior is what he had in mind.

> I think this is probably clear enough that we should just go ahead and do it. The Mac "Delete" key fires a Backspace event, do I have that right?

Yeah maybe we can just push this through. It's definitely a Backspace event on Mac keyboards, confirmed via https://keycode.info.

simonwiles:
Thanks -- just one problem as I go to implement this quickly -- what to put on the button?  <kbd>Del</kbd> ?  <kbd>Backspace</kbd> ? Something else??!!

broadwell:
I guess <kbd>Backspace</kbd> is the best/least worst option, especially thinking that this will need to coexist with the ability to choose custom key bindings, where any deviation from the official key code conventions is going to be a pita down the line. Maybe we could explain somewhere in the documentation that <kbd>Backspace</kbd> = <kbd>delete</kbd> on Mac systems.

simonwiles:
> I guess Backspace is the best/least worst option, especially thinking that this will need to coexist with the ability to choose custom key bindings, where any deviation from the official key code conventions is going to be a pita down the line. Maybe we could explain somewhere in the documentation that Backspace = delete on Mac systems.

We'll go with that for now, then, but the text doesn't have to correspond with the `key`, `keyCode`, or `code` properties of the keyboard events.

broadwell:
Yeah I can see that we may end up using custom names for some of the keys, to save space in the UI if for no other reason... but maybe we can avoid that for now.

PR #95, opened 2021-08-28
Refactor `Piano` to allow updating sample volumes, velocities, and reverb

simonwiles:
This is a PR against your `reverb` branch which should present the same user-facing behaviour, just with a few backstage optimizations and reorganizations.

Changing the number of velocities should be robust, and quicker since the whole `Piano` doesn't have to be reloaded, but the next step should still be to have a loader of some kind when the samples are loading, and probably an "are you sure?" dialog when trying to go up from the default (with a "don't ask me again" checkbox, I think).

I wanted to do this on top of your branch, rather than replacing it, and then I decided to PR this bit now so that you could easily see the changes thus far.  If you're okay with merging this in to `reverb`, I'll carry on adding commits there to address the UX (including the above two points and probably something for the `AudioSettings` panel too).  Hope that's cool.

### DIFF
--- a/src/pianolatron-piano.js
+++ b/src/pianolatron-piano.js
@@ -26,6 +26,7 @@ export class Piano extends ToneJsPiano {
         samples: this.options.url,
         volume: this.options.volume.strings,
       }).connect(this.output);
+      this.strings = this._strings.volume;
       return this._strings.load();
     }
     // eslint-disable-next-line prefer-promise-reject-errors


broadwell:
OK thanks, this looks like it's definitely on the right track. Just to make things more complicated, I decided to merge in some rebasing work I had already done to the `key-velocities-map` branch (which is also based on `reverb`) to bring it up to date with `main` and particularly with the use of the `<Icon>` component. I assume you'll be overhauling much of that anyway, but anyway it's in now and  hopefully didn't do any harm.

Btw there's no need to worry about getting out of sync with the `key-velocities-map` branch after this -- I can always merge these commits into that one later, if needed, but it's basically just a demo branch and so may not see much further development.

broadwell:
Oh actually I do notice one issue (hopefully not caused by my risky merge above): changes to the  "Strings" value in the Piano Sample Volumes menu have no effect if the number of samples has previously been changed. Changing the volumes of the other sample types however (Harmonics, Pedals, Keybed) continues to have noticeable effects (though compared to the Strings, they're so quiet that it's hard to notice).
Presumably the `this.strings.value` accessed in `piano.updateVolumes()` needs to be re-connected somehow to the new `this._strings` after it's recycled in `piano.updateVelocities()`?

There are also some linter errors due to bits of the old `reverb` code that have been made redundant, but I'm sure those will get closed out later.

simonwiles:
> There are also some linter errors due to bits of the old reverb code that have been made redundant, but I'm sure those will get closed out later.

Yeah, the rebase has gone wrong in places and made a bit of a mess.  That's where these have come from.

> Oh actually I do notice one issue (hopefully not caused by my risky merge above): changes to the "Strings" value in the Piano Sample Volumes menu have no effect if the number of samples has previously been changed. Changing the volumes of the other sample types however (Harmonics, Pedals, Keybed) continues to have noticeable effects (though compared to the Strings, they're so quiet that it's hard to notice).
Presumably the this.strings.value accessed in piano.updateVolumes() needs to be re-connected somehow to the new this._strings after it's recycled in piano.updateVelocities()?

Thanks, that's a good catch :)

I'm going to redo the rebase and fix this error quickly, and then I'm going to push the whole lot over the top of the `reverb` branch, and close this PR.

simonwiles:
Huh, that's interesting -- I rebased everything correctly and GitHub recognized that the force-push to `reverb` actually resolved this PR perfectly, so it marked it as merged automatically.  That's pretty clever.

PR #96, opened 2021-08-30
Three tiny fixes

simonwiles:


broadwell:
This all looks good to me. Now the only debug warning I'm seeing (besides the warning about ToneJS using `ScriptProcessorNode`) is
`<App> received an unexpected slot "default".`

And fortunately I hadn't noticed the persisted store reset bug at all.

simonwiles:
Thank you!

> <App> received an unexpected slot "default".

This is a result of interaction between Svelte and Astro -- don't think there's anything I can do about (short of something upstream in Astro).  Like the other debug messages, though, it's really just a polite warning that only appears in dev. mode.

> And fortunately I hadn't noticed the persisted store reset bug at all.

Yeah, it's a theoretical thing, really.  I'm not entirely convinced of the wisdom of the `createStore` function that adds a `reset` method anymore, but I carried it across into `createPersistentStore`, and so even though we're not using that code anywhere (hence you didn't notice it didn't work), it ought to be fixed so as to reduce the WTF count if we ever do try to use it in the future!

PR #97, opened 2021-08-30
Housekeeping

simonwiles:
This PR is mostly dependency bumps (want to keep up with Astro in particular), but also bumps all the linting and formatting dependencies, and updates a couple of source code files as a direct result.  Finally, this PR also includes the addition of a couple of new `<meta/>` tags in the `<head/>` of rendered pages, stamping in the build date and time and the git commit the deploy was built from.  Often handy.

Will need a quick `rm -rf node_modules && yarn install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,16 @@ Development repository for the CIDR Pianolatron application.
 ### Live Updating Development Build
 
 1. `yarn`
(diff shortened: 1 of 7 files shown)


broadwell:
No problems with these changes, other than the issue of the `astro preview` site not handling the `?druid=` URL param in my setup. But that's a minor concern.

PR #98, opened 2021-08-31
Add scalebar to viewer

broadwell:
Feedback requested. It's been bothering me that there's really no context in the app to indicate how big or small the rolls are, other than the scan reference card way at the end of the ribbon. This is just an ES6-ified version of the NIST's [OpenSeadragon scalebar](https://github.com/usnistgov/OpenSeadragonScalebar), and it can be dropped in with a minimum of fuss.

Something a bit more ruler-like in appearance and function might be preferable, but this works well enough for me. It has the option of displaying metric units, but if we'll also be exposing the speed and acceleration rates in feet, then perhaps we should stick with imperial (or better yet, allow for selecting one or the other).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 src/tonejs-piano.js
+src/scalebar.js
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -86,12 +86,7 @@
   import { onMount } from "svelte";
   import { fade } from "svelte/transition";
   import OpenSeadragon from "openseadragon";
-  import {
-    Scalebar,
-    ScalebarType,
-    ScalebarLocation,
-    ScalebarSizeAndTextRenderer,
-  } from "../scalebar";
+  import { Scalebar } from "../scalebar";
   import {
     rollMetadata,
     scrollDownwards,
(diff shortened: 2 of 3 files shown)


broadwell:
Suggestions for improved styling would be great, too, if this implementation is generally acceptable.

simonwiles:
I think this is a great idea.  I'm happy to have a crack at styling it up a bit. 

I sort of feel like this is a bit of an ott approach, though, if I'm not being too much of a pain.  Ours is a pretty simple case -- could we just use something like this: https://github.com/sul-cidr/pianolatron/commit/6a10f8edb12aafd1557b51107f1ce947d3474396 ?  (styling still to come, of course...)

> It has the option of displaying metric units, but if we'll also be exposing the speed and acceleration rates in feet, then perhaps we should stick with imperial (or better yet, allow for selecting one or the other).

I've had in mind a metric/US conversion button somewhere -- US by default, but with the option to check a box and have all the appropriate places converted.  Doesn't seem like too much work.

simonwiles:
Lol at the bug whereby zooming out with the mouse wheel is never capped:

image

I think if I've zoomed out to 500,000 miles, I may have gone too far...

 (Fixed in the branch referenced above, can be cherry-picked easily onto another branch as desired.)

simonwiles:
The scale bar could be added to the "legend" panel too, tbh -- partly so as to be able to turn it on and off without the need for yet another checkbox.

broadwell:
Yeah, I'd be fine with using `alt-scalebar` instead (with a bit more styling). We'll be out of luck if we find ourselves working with a planetary-sized piano roll, however.

I also just kinda wanted to update the NIST scalebar because I've encountered it in a few different OSD/Mirador-related contexts over the years, and it has always bothered me that there isn't a more up-to-date standard scalebar implementation. Oh well.

I'm having some trouble envisioning how the bar would fit into the legend panel, however, although I agree it's preferable to minimize the number of checkboxes.

broadwell:
Closed in favor of #100

PR #99, opened 2021-09-01
Reorganize Config

simonwiles:
The initial impetus behind this branch was to reorganize the keyMap config so as to make a fix for the behaviour whereby the editing state doesn't reset when the "Reset to defaults" button was clicked (and, relatedly, that it is possible to have multiple shortcuts in the "Choose a new keystroke..." editing state, which is even more egregious).  The fixes ended up on another branch, and this one turned in to a general reorg of config, library, and stores code.


src/config/keyboard-shortcut-config.js
```diff
@@ -0,0 +1,125 @@
...
+    description: "Pan Downwards",
+    help: "Pan the roll downwards (hold to accelerate)",
+  },
+};
+
+export const unusableKeys = [
+  "Escape",
+  "ControlLeft",
+  "ControlRight",
+  "Enter",
```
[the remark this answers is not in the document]
simonwiles: Yeah, I'm down with that -- not sure why I put it on that list in the first place, tbh.

src/config/keyboard-shortcut-config.js
```diff
@@ -0,0 +1,124 @@
...
+  "ShiftLeft",
+  "ShiftRight",
+  "CapsLock",
+];
+
+export const alternativeIndicatorText = {
+  Space: "＿",
+  Backspace: "⌫",
+  Delete: "⌦",
+  NumpadEnter: "⏎",
```
broadwell, line 119: Cool, now that <kbd>Enter</kbd> is allowed, should we add the same alternative indicator symbol for it here (⏎)? That could lead to some confusion if the user decides to remap both Enter keys (assuming they're using a keyboard that has both), but that's a pretty extreme corner case, and it wouldn't really affect the app's overall functionality anyway, so I'm OK with it.

simonwiles: Yes, I suppose so, thanks.  It's the same situation as numbers on the `Numpad` or on the top row -- either will show, e.g. <kbd>7</kbd>, but only one will actually work...  Increasingly uncomfortable with it, tbh, but this will be okay for now, I guess.

simonwiles:
> The granularity of the sliders (including the tempo slider) is sometimes a bit of a surprise initially, but conceptually I think it makes sense for it to correspond to the default granularity on keyboard input.

Yeah, I'm willing to move on this if desired (e.g. perhaps the increment/decrement on key-press should be smaller), but the consistency is valuable in-and-of itself, imo.  Thanks!

PR #100, opened 2021-09-01
Alternative scalebar

simonwiles:
I'm going to open this as a draft PR now.  It has some basic ruler-like styling (the `.us` class is there as a place-holder as this will need to be tweaked for a metric style).  At the moment it fades in where the "Downloading roll images..." fades out, which doesn't seem awful but I'm not wedded to it either.

Open season on this at this point -- how far is it from something you'd be happy with?  Really we could do anything we want with it, so I'd like to hear your wishlist.  A solution is certainly needed for the ticks when we're at multiples (especially odd multiples) of an inch.  Some other possibilities include setting a maximum width, and optimizing the number of "ticks" shown depending on the width.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -162,6 +162,7 @@
   };
 
   const resetApp = () => {
+    rollViewer?.$destroy();
     mididataReady = false;
     clearNotification();
     appReady = false;
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -6,17 +6,6 @@
     height: 100%;
     width: 100%;
 
-    p {
-      background: rgba(black, 0.4);
-      border-radius: 4px;
-      color: white;
-      left: 1em;
-      padding: 4px 8px;
-      position: absolute;
-      top: 1em;
-      z-index: 1;
-    }
-
     // tracker bar
     &::before {
       background: linear-gradient(
@@ -81,13 +70,18 @@
     }
   }
 
-  .scale-bar {
+  span {
+    background: rgba(black, 0.4);
+    border-radius: 4px;
+    color: white;
+    left: 1em;
+    padding: 4px 8px;
     position: absolute;
-    height: 20px;
-    display: block;
-    z-index: 10;
+    top: 1em;
+    z-index: 1;
+  }
+  .scale-bar {
     background: rgba(grey, 0.8);
-    padding: 0 1em 0;
   }
 </style>
 
@@ -539,7 +533,13 @@
   style={`--trackerbar-height: ${trackerbarHeight}px;`}
 >
   {#if !rollImageReady}
-    <p transition:fade>Downloading roll image...</p>
+    <span class="roll-loading" transition:fade>Downloading roll image...</span>
+  {:else}
+    <span
+      class="scale-bar"
+      transition:fade
+      style={`width: ${scaleBarWidth.px}px`}>{scaleBarWidth.label}</span
+    >
   {/if}
   {#if showControls}
     <RollViewerControls
@@ -549,7 +549,4 @@
       {updateTickByViewportIncrement}
     />
   {/if}
-  <span class="scale-bar" style={`width: ${scaleBarWidth.px}px`}
-    >{scaleBarWidth.label}</span
-  >
 </div>
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -81,7 +81,27 @@
     z-index: 1;
   }
   .scale-bar {
-    background: rgba(grey, 0.8);
+    background-color: rgba(white, 0.5);
+    background-repeat: no-repeat;
+    border-radius: 0;
+    border-top: 0;
+    border: 1px solid black;
+    color: black;
+    font-size: 12px;
+    height: 22px;
+    padding: 0 2px;
+    transition: width 0.5s ease;
+
+    &.us {
+      background-image: linear-gradient(black, black),
+        linear-gradient(black, black), linear-gradient(black, black),
+        linear-gradient(black, black), linear-gradient(black, black),
+        linear-gradient(black, black), linear-gradient(black, black);
+      background-position: 12.5% bottom, 25% bottom, 37.5% bottom, 50% bottom,
+        62.5% bottom, 75% bottom, 87.5% bottom;
+      background-size: 1px 6px, 1px 9px, 1px 6px, 1px 12px, 1px 6px, 1px 9px,
+        1px 6px;
+    }
   }
 </style>
 
@@ -536,7 +556,7 @@
     <span class="roll-loading" transition:fade>Downloading roll image...</span>
   {:else}
     <span
-      class="scale-bar"
+      class="scale-bar us"
       transition:fade
       style={`width: ${scaleBarWidth.px}px`}>{scaleBarWidth.label}</span
     >
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -80,63 +80,6 @@
     top: 1em;
     z-index: 1;
   }
-
-  .scale-bar {
-    background-color: rgba(white, 0.5);
-    color: black;
-    font-size: 12px;
-    height: 22px;
-    transition: width 0.5s ease;
-    position: absolute;
-    left: 1em;
-    top: 1em;
-    z-index: 1;
-    display: flex;
-    border: 1px solid black;
-    border-top: 0;
-    border-right: 0;
-
-    &:before {
-      content: attr(data-label);
-      display: block;
-      right: 4px;
-      position: absolute;
-      top: calc(100% + 2px);
-    }
-
-    span {
-      background-repeat: no-repeat;
-      border-radius: 0;
-      flex: 1 0 auto;
-    }
-
-    &.us span {
-      background-image: linear-gradient(black, black),
-        linear-gradient(black, black);
-      background-position: 50% bottom, 100% bottom;
-      background-size: 1px 12px, 1px 22px;
-    }
-
-    &.us.major span {
-      background-image: linear-gradient(black, black),
-        linear-gradient(black, black), linear-gradient(black, black),
-        linear-gradient(black, black);
-      background-position: 25% bottom, 50% bottom, 75% bottom, 100% bottom;
-      background-size: 1px 9px, 1px 12px, 1px 9px, 1px 22px;
-    }
-
-    &.us.minor span {
-      background-image: linear-gradient(black, black),
-        linear-gradient(black, black), linear-gradient(black, black),
-        linear-gradient(black, black), linear-gradient(black, black),
-        linear-gradient(black, black), linear-gradient(black, black),
-        linear-gradient(black, black);
-      background-position: 12.5% bottom, 25% bottom, 37.5% bottom, 50% bottom,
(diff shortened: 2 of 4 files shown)


simonwiles:
Added a possible solution to the multiples problem (though now it's too cramped when zoomed well out).

simonwiles:
Right, I promise I'll stop messing around with this now, and let you look at it before I make it dance around any more :)

broadwell:
Making the bar more ruler-like is definitely preferable! The only caveat is that if it looks like a ruler, people will want to use it as one, so it really needs to be accurate -- and OSD's animation behavior makes this trickier than you'd think. Specifically, just resizing on `zoom` isn't sufficient -- note that the width of the two adjacent hole columns as measured with the ruler seems to vary between zoom levels:
Ruler_with_zoom
Following the lead of the NIST scalebar code, I updated the scalebar to resize on `animation` and `resize` instead, which irons out the ruler's measurements at different zoom levels:
Ruler_with_animation_and_resize
Rescaling on `resize` is maybe not strictly necessary because we don't let the zoom of the viewer change on window resize, but it's probably better to be safe than sorry. In any case, the long delay before it snaps to the correct proportions after zooming isn't ideal, but I'm not sure what we can do about that.
Making this change this for the trackerbar resizing is also probably warranted and seems to squelch the bouncing behavior we've noticed, so all together the code would be something like this:
```
    // update the height of the tracker bar when the zoom changes
    const scaleTrackerbar = () => {
      const imageZoom = viewport.viewportToImageZoom(viewport.getZoom(true));
      trackerbarHeight = Math.max(1, avgHoleWidth * imageZoom);
    };

    const scaleScalebar = () => {
      const minScaleBarWidth = 100;
      const ppi = viewport.viewportToImageZoom(viewport.getZoom(true)) * 300;
      let inches = 1;
      while (ppi * inches < minScaleBarWidth && inches < 12) inches += 1;
      scaleBarWidth = {
        px: ppi * inches,
        label: `${inches}in`,
        multiples: inches,
      };
    };

    openSeadragon.addHandler("animation", () => {
      scaleTrackerbar();
      scaleScalebar();
    });

    openSeadragon.addHandler("resize", () => {
      scaleTrackerbar();
      scaleScalebar();
    });
```

broadwell:
Some other, more superficial comments:
- It's a little hard to read the numbers against a dark background -- maybe the entire scalebar should have a light semitransparent mask behind it?
- Maybe cap the ruler length at 1ft (and thin out the ticks when zoomed out, as you've mentioned)? Not sure how this would coexist with the metric support, which also seems to be on the menu.

simonwiles:
> It's a little hard to read the numbers against a dark background -- maybe the entire scalebar should have a light semitransparent mask behind it?

Yeah, I tried a couple of things and couldn't find anything I liked at all.  Will have another look.  Agree that something's needed.

> Maybe cap the ruler length at 1ft (and thin out the ticks when zoomed out, as you've mentioned)? Not sure how this would coexist with the metric support, which also seems to be on the menu.

I'll take a look at this, too.  Automatic adjusting of ticks should be implemented, bu might need more work at the two extremes, though?  Supporting metric shouldn't be a problem, I think.

broadwell:
Using `viewport.getZoom(false)` gets the job done, and also seems preferable to the proposed fix in that it doesn't involve that bizarrely long pause after the animation before the scalebar snaps to the proper size.

And probably we don't need to worry about handing `resize` events as long as the viewer is created with `preserveImageSizeOnResize: true`

simonwiles:
Thanks!  Yeah, it really should only be necessary to change the scale bar when the, uh, scale changes... :)

simonwiles:
I've tweaked this a little (including making it draggable), and I'm marking it "ready for review" now.  I've a nagging feeling I've forgotten something, though -- is there anything outstanding here?

simonwiles:
So I couldn't leave it well alone -- there's an alternative style in the `alt-scalebar-update` branch:
Peek 2021-09-02 10-54

Let me know if you like either of these more than the other.

broadwell:
Cool, overall I like the styling in `alt-scalebar-update` a bit better -- I think it's preferable not to have the "base" unit of the bar (1in, 2in) constantly changing, and only available in the label at bottom right. The numbers in the updated version do get a bit crowded when zoomed all the way out, but I can live with that.
Screen Shot 2021-09-02 at 11 32 51 AM

Making the bar draggable is great, although being able to move it outside of the viewer window feels a bit odd, and can lead to some perplexing UI experiences -- the most notable being that it's possible to "lose" the scalebar off the edge of the browser window entirely, and to lose it under the keyboard graphics. Bumping the scalebar's z-index, if possible, would address the latter, but it might be preferable just to corral the ruler within the OSD window, if possible.

Being able to drag the bar around also increases the desirability of enabling rotation by 90 degrees, which was already sort of present before (vertical distances often being of greater interest). I think in that case it would ideally rotate 90 degrees counterclockwise, with the "inches" label maybe shifting so it's always below the entire bar vertically.

simonwiles:
> The numbers in the updated version do get a bit crowded when zoomed all the way out, but I can live with that.

Yeah, not sure how much more can realistically be done about that at this point.

> Making the bar draggable is great, although being able to move it outside of the viewer window feels a bit odd, and can lead to some perplexing UI experiences -- the most notable being that it's possible to "lose" the scalebar off the edge of the browser window entirely, and to lose it under the keyboard graphics. Bumping the scalebar's z-index, if possible, would address the latter, but it might be preferable just to corral the ruler within the OSD window, if possible.

I wondered about this, but thought I could get away without implementing it (partly on the grounds that no matter how badly you manage to lose it, it pops back to it's original position on page reload or loading a new roll -- if, indeed, the new roll will load at all...).  But corralling is probably worth the effort, I suppose :)

> Being able to drag the bar around also increases the desirability of enabling rotation by 90 degrees, which was already sort of present before (vertical distances often being of greater interest). I think in that case it would ideally rotate 90 degrees counterclockwise, with the "inches" label maybe shifting so it's always below the entire bar vertically.

Hmm, now I didn't think of this, but you're right of course.  Rotating through 90⁰ should be straightforward enough, although it brings to the fore another gripe (I did wonder whether I should have enabled draggability and opened this can of worms...!): you kinda really want to be able to stretch it, right?  Does enabling all this wind up for a somehow worse UX if this isn't possible too?

broadwell:
> I wondered about this, but thought I could get away without implementing it (partly on the grounds that no matter how badly you manage to lose it, it pops back to it's original position on page reload or loading a new roll -- if, indeed, the new roll will load at all...). But corralling is probably worth the effort, I suppose :)

Yes, it's more that being able to lose the bar, or even put it somewhere it doesn't really have any reason to go, like over the metadata sidebar, feels kind of out of character for the app, which otherwise puts nice comfortable guardrails around all aspects of the UX.

> Hmm, now I didn't think of this, but you're right of course. Rotating through 90⁰ should be straightforward enough, although it brings to the fore another gripe (I did wonder whether I should have enabled draggability and opened this can of worms...!): you kinda really want to be able to stretch it, right? Does enabling all this wind up for a somehow worse UX if this isn't possible too?

It is a bit of a slippery slope, for sure, but I feel that being able to move the bar around (and maybe rotate it) and then zoom in and out to measure at different scales does sort of lessen the immediate appetite for stretchability. If we were to provide that kind of functionality, then probably it would be even more desirable (and maybe the same amount of work) to implement a "measuring mode" wherein the user can just click or click and drag to create a line/path, and the length of that path in inches is displayed somewhere.

simonwiles:
> Yes, it's more that being able to lose the bar, or even put it somewhere it doesn't really have any reason to go, like over the metadata sidebar, feels kind of out of character for the app, which otherwise puts nice comfortable guardrails around all aspects of the UX.

> It is a bit of a slippery slope, for sure, but I feel that being able to move the bar around (and maybe rotate it) and then zoom in and out to measure at different scales does sort of lessen the immediate appetite for stretchability.

Agreed on both counts.  Commits incoming (although maybe this evening now).

> If we were to provide that kind of functionality, then probably it would be even more desirable (and maybe the same amount of work) to implement a "measuring mode" wherein the user can just click or click and drag to create a line/path, and the length of that path in inches is displayed somewhere.

Yeah, this is where my mind was going.  Slightly surprised to see there doesn't appear to be an OSD plugin that does this already.

simonwiles:
I think this is rebased and ready for review with a view to merging at this point?

broadwell (changes requested):
After the rebase, switching between rolls no longer works, presumably because the persistence of the un-destroyed `<RollViewerScalebar />` element prevents the viewer from being recycled when `loadRoll()` -> `resetApp()` sets `appReady` to `false.`

Once that's fixed, I think the PR will be ready to go in.

Support for metric units can come later, probably in concert with the display of velocity and maybe acceleration measurements via the legend widget or elsewhere.

It's possible for the scalebar to get stuck under the piano keyboard overlay (when it's drawn over the viewer), but hiding or popping out the keyboard makes it accessible again, so it's OK.

Also, I was thinking that we might not want to allow the ruler to move over the progress bar/navigator, but letting it go there actually makes it easier to measure items on the right edge of the viewer (when the ruler is in vertical orientation), and no one would reasonably assume that the ruler's measurements apply to the navigator view of the roll, so this is fine.

simonwiles:
Ugh, sorry, mea culpa 😳

simonwiles:
Right -- I've been on a bit of a deep-dive on this, and it seems the problem is a Svelte bug (see the referenced issue on the Svelte repo).  In the process of trying to address it I've tried a number of solutions, but I think that f22784b, while a little kludgey, should prevent the problem from arising.

Essentially the `<RollViewer/>` component wasn't being cleaned up if it had a child element or component with a transition at the moment when the `<RollViewer/>` component was removed from the DOM.  This could happen prior to this PR when the roll was changed while `<RollViewerControls/>` were shown (or still fading out), or perhaps if the "Downloading roll image..." was still showing.

[Does this seem like it corresponds to the behaviour as you perceived it?  Now that I'm aware that this is a thing I can confirm that behaviour, but you were always able to produce this error more easily than I -- perhaps there are still other circumstances?]

Anyway, with the introduction of the `<RollViewerScaleBar/>` this situation is guaranteed to obtain at all times, so the bug happens on every roll change event.

Hopefully this is permanently fixed now.

broadwell:
Aha! This is a great find. The description of the bug tracks very well with the behavior we've been seeing, including the more or less random occurrences prior to the introduction of the `<RollViewerScaleBar/>`. It's a pity it hasn't been fixed yet, but [f22784b](https://github.com/sul-cidr/pianolatron/commit/f22784b711222e768c86f76eccb4b98942d937c5) seems to be an appropriate workaround.

I've been reloading rolls for the past half hour or so, in both Firefox and Chrome, and they've all loaded fine, so if the bug is still lingering in some other form, its profile has at least been greatly reduced. I think the branch is ready to merge now.

broadwell:
Btw, there's some noticeable lag when dragging the ruler, but only for certain rolls and browsers; specifically, it's most obvious with MacOS + Chrome + long/complicated rolls, and nearly absent for MacOS + Firefox + any type of roll.
Sticky_scalebar
It's a little easier to see how this ties into the general jerkiness of the viewer when the dragging is done during playback, although this is a fairly contrived scenario:
Sticky_scalebar_playing
Just mentioning it now because it's likely also a consequence of the OS + browser combo not coping well with large SVG overlays on the OSD viewer. So any refinements to alleviate that situation likely will help with this "sticky ruler" behavior, too.

PR #101, opened 2021-09-03
Welcome Splash Screen

simonwiles:
I feel like we need something for first-time visitors.  We'll want to get the PIs' input on what should actually be said (and/or linked) here, but feel free to change / suggest things that should be changed before it even goes in.

PR also includes a "Reset All Settings" button, which reactivates the splash screen as well as resetting the UI-related settings (basically everything in `$userSettings` -- which now includes the piano keyboard state -- but not the (computer) keyboard bindings).

PRing now as it can be looked at and potentially merged before the other ongoing things will be ready.


src/Pianolatron.svelte
```diff
@@ -291,11 +291,11 @@
...
-  {#if $showKeyboard && !$overlayKeyboard}
+  {#if $userSettings.showKeyboard && !$userSettings.overlayKeyboard}
     <div id="keyboard-container" transition:slide>
       <Keyboard keyCount="88" {activeNotes} {startNote} {stopNote} />
     </div>
-  {:else if !$showKeyboard}
+  {:else if !$userSettings.showKeyboard}
     <KeyboardControls outside />
   {/if}
   {#if !appReady}
```
broadwell, line 301: This is only sort of relevant (loading messages are tangentially connected to the welcome and notification UX), but it has bothered me that the spacing between the throbber bars changes when the message changes from "Loading application" to "Loading resources". This filler formula seems to make both of them use the same spacing:
`<div><span />&nbsp;<span />&nbsp;<span />&nbsp;<span />&nbsp;<span /></div>`

broadwell: In order for the spacing to be consistent, `index.astro` should be modified to use the same string, I think.

broadwell:
A technical issue: I like the new notification formattin, but as I recall it used to be possible to trigger a notification by entering an invalid DRUID value in the URL, and now this doesn't seem to work.

simonwiles:
> A technical issue: I like the new notification formatting, but as I recall it used to be possible to trigger a notification by entering an invalid DRUID value in the URL, and now this doesn't seem to work.

Yeah that's actually broken in production right now (I only noticed when working on this PR, though).  It's on my list.

src/Pianolatron.svelte
simonwiles: Huh, yeah, thanks.  This actually got seriously broken at one point and I thought I'd fixed it, but evidently I only fixed it about 90%.  There's still a discontinuity when the app code is finished parsing as the animation restarts, and I don't love that.  I'd like a slightly different solution to the whole thing, but I don't think it's a high enough priority right now.

broadwell:
I'm not fully convinced that a welcome splash screen is necessary, but this one looks nice, and I don't mind adding it now so that the PIs can see what's possible and then decide whether they want one or not.

It seems that the contents of the welcome screen would mostly just be links to documentation elsewhere within the app (how to use, about this app, etc.), whose availability probably also would be apparent from the presence of a menu bar or at least a hamburger of some kind somewhere in the top-level app UI.

I also think that a welcome splash screen might offer an unwelcome interruption to the process of opening a roll via a link from elsewhere -- maybe even first-time users really will just want to get to the playback as quickly as possible. Though an optional URL parameter that suppresses the splash screen would be one way to handle this.

This also does bring up the tangential issue of logos and branding, i.e., which logos should the splash screen or other docs include (SUL, CIDR, CCRMA); neither the library's Player Piano Program nor SUPRA actually has a decent logo, either...

simonwiles:
Yeah, these are all fair comments.  I feel like first-time visitors need something, rather than just a random roll.  If you click the button you'll literally just see it once.

> I also think that a welcome splash screen might offer an unwelcome interruption to the process of opening a roll via a link from elsewhere -- maybe even first-time users really will just want to get to the playback as quickly as possible. Though an optional URL parameter that suppresses the splash screen would be one way to handle this.

My assumption is that we'd have this turned off if embedded or otherwise integrated.  A URL param would be one way of doing this, yeah.

broadwell:
Forgot to bring this up during the meeting today, but anyway I think that once it's rebased (and verified not to trigger the roll-switching error), it's fine to push this up so that the PIs can see it.

broadwell:
I'm not necessarily advocating this, but we could sniff the user agent and display caveats on the splash screen about specific browser + OS combos if detected, since IE and Safari are both more or less unsupported at this point. Such messages can quickly get out of date and/or become inaccurate though, as with Orbis V2. But even if we don't do an actual browser/OS check, it might be appropriate to put a blurb on the splash screen, something like

"Some devices and web browsers may not be able to run the Pianolatron with all features enabled. Please see the documentation."

simonwiles:
Rebased (and merge conflicts resolved) so that this can go straight in if #103 goes in as-is first.

I'm in favour of some "warning" information on the welcome panel, though not immediately clear on what form I think it should take.  I'll note here for now that the js we're shipping won't even be downloaded and parsed on IE, so an alternative method of displaying a notice would be required there.

broadwell:
OK cool, this works for me. I think it might be somewhat preferable if the welcome screen acted like a full-focus modal, i.e., `data-backdrop="static"` in Bootstrap land (not sure how easy/difficult this is in Svelte-land), so that any clicks outside of it do nothing. Being able to click around the interface and even drag things under the message box is a bit weird. But this is fine as a proof-of-concept to elicit feedback from the PIs.

simonwiles:
> I think it might be somewhat preferable if the welcome screen acted like a full-focus modal, i.e., data-backdrop="static" in Bootstrap land (not sure how easy/difficult this is in Svelte-land), so that any clicks outside of it do nothing. Being able to click around the interface and even drag things under the message box is a bit weird.

Yeah, this could be done.  I kinda deliberately didn't, in the first instance, because I didn't want it to feel like it was blocking the app, like it was enforcing some kind of contract that you had to agree to to be permitted access somehow.  Happy to add this, though (it should be very simple).  It could also just be that the panel is dismissed by a click anywhere on the page?

broadwell:
I think that clicking anywhere on the page to dismiss would be great -- forcing people to hunt down the "Hide forever" button probably isn't a good look, especially if it may be difficult to reach on certain devices.

simonwiles:
Well, you can click anywhere on the panel or hit <kbd>Esc</kbd> to dismiss right now (not permanently, though).  But I guess it could just be "click anywhere" -- I'll implement that.

broadwell:
Hmm right, I guess it wasn't super obvious to me that clicking anywhere other than the "dismiss permanently" button was equivalent to a "dismiss temporarily". It seems like the prevailing UI for this kind of thing is a checkbox with "Don't show me this again" next to it, and then a generic "close" button (so the checkbox gives visual reinforcement to the idea that the choice is persistent).

But anyway, that's a minor quibble; I think the bigger question is what the welcome panel will be used for; if it will include links to other documentation, then clicking anywhere on the panel to dismiss it probably won't fly, although clicking anywhere *outside* the panel would still be fine. But maybe it won't be used that way at all. Probably the best thing for now is just to see if we can get any feedback out of the PIs about it.

PR #102, opened 2021-09-07
Fixes and updates

simonwiles:
This PR includes:

1. a fix for mousewheel zooming (this one has been on a couple of branches but hasn't gone in yet -- it should go in here)
2. a fix for the race condition whereby `annotateHoleData()` could be called before the hole data was available (which is blocking deployment of #101 and #100, I think)
3. a fix for the notification not showing when an invalid DRUID is supplied via http get params, which was introduced in #93 

Along with (3), the notification itself has been tweaked and access to the roll selector widget has been made available, so that rather than just having to reload the page without the defective DRUID param, a user can just pick a roll directly from that page load.

(2) was originally intended to be merged along with a more thorough overhaul of the way the hole data is passed around the application in concert with the expressionizer implementation.  The commits here are just the first step of that, but if they go in now we can deploy the other code that's waiting on this part.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,10 +17,11 @@ module.exports = {
     },
   ],
   rules: {
-    "no-underscore-dangle": "off",
-    "no-return-assign": ["error", "except-parens"],
     "no-cond-assign": ["error", "except-parens"],
     "no-param-reassign": ["error", { props: false }],
+    "no-return-assign": ["error", "except-parens"],
+    "no-undef-init": "off",
+    "no-underscore-dangle": "off",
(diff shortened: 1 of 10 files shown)


broadwell:
Just FYI, the race condition on roll load is still around after these changes; it just no longer causes the app to hang entirely upon finding an empty `holesData` in `annotateHoleData()`. Instead, it loads the JSON (metadata) and MIDI for the new roll, but continues to display (and scroll) the old roll in the viewer. But as the description indicates, this is an intermediate step towards a more comprehensive solution, and it definitely improves on the current behavior, so I think it's good to go in.

PR #103, opened 2021-09-16
Better tooltips

simonwiles:
This PR replaces the simple CSS-based tooltips with something that depends on Javascript instead.

The primary motivation for this is the need to anchor the tooltip elements to the highest part of the DOM tree as possible (`document.body` in this case) and then position them appropriately, so as to alleviate clipping context issues.

Right now the behaviour is like-for-like (with the above exception), but this will allow other improvements such as HTML tooltips, automatic flipping and positioning, and overflow-prevention to be added very easily if/when we need them.

In a departure from my usual MO, this functionality leans on popper to do all the heavy lifting (so a `yarn install` will be required after checking out this branch).  For this very simple application only the smallest amount of popper's range of functionality is required; for this reason the minimum amount of library code is imported, but it still adds ~3.5K to the shipped bundle.  However, the point is to be as close to bullet-proof as possible and not spend ages chasing edge cases, and popper as a positioning library is a solid as it gets in this respect.  Adding tooltips to other parts of the application should be more straightforward this way.


src/components/KeyboardShortcutEditorRow.svelte
```diff
@@ -51,22 +52,22 @@
...
-    data-tooltip="Edit"
+    use:tooltip={"Edit"}
     bind:this={editButtonRef}
     on:click={() => (editing = !editing)}
   >
     <Icon name="edit" height="20" width="20" />
   </button>
   <button
-    data-tooltip={shortcut.isChanged ? "Reset to Default" : undefined}
+    use:tooltip={shortcut.isChanged ? "Reset to Default" : undefined}
```
broadwell, line 70: These props seem to be handled differently under the Popper regime. Here's what the tooltip looks like for the reset button on a key binding that has never been changed from the default:
Screen Shot 2021-09-16 at 12 03 12 PM
Note also that the "Reset to Default" text persists for a binding that has been reset, until the Keyboard Controls popup is closed and reopened:
Screen Shot 2021-09-16 at 12 04 44 PM
In this case, the simplest fix is just to remove the tooltip entirely, since it should be pretty obvious what the button does. But a more systematic fix probably involves some tweaks to the tooltip code.

broadwell (changes requested):
I'm totally cool with using Popper for this, and the new tips look good. Probably we should address the issue with the reset buttons before merging, though.

src/components/KeyboardShortcutEditorRow.svelte
broadwell: And yikes, the behavior is different on Opera (Mac OS). The `undefined` value for the `text` parameter results in an absent tooltip, which is fine, but once the key has been remapped, the tooltip looks like this:
Screen Shot 2021-09-16 at 12 14 06 PM
and as above, the "undefined" tooltip hangs around after the reset button has been clicked.

simonwiles: I've not been able to reproduce either of these behaviours exactly (presumable for want of a Mac, and Opera especially), but I've just pushed some updates that improve the way the tooltips are bound, and hopefully they'll address this.  Behaviour is solid on FF, Chromium, and Opera on Linux and Windows 10 in VM -- could you give it a spin for me?

broadwell:
Great, these updates fix the tooltip behavior on Mac OS Firefox and Chrome. I think this is ready to merge.

PR #104, opened 2021-09-16
Split SVG holes overlay into partitions, only display sections that overlap with viewer extent

broadwell:
This change has been confirmed to alleviate the playback slowdowns experienced on relatively recent Mac OS machines running Chromium browsers when playing rolls with complicated overlays (i.e., lots of holes). Firefox on Mac OS does not exhibit this difficulty in shifting large SVGs (at least on the machines available for testing), so it's not affected.

It should be noted that unfortunately, performance issues persist on other platforms, regardless of whether or not the roll has a large SVG overlay: Mac OS Safari, browsers of any type on middling Windows systems, and non-premium Android and iOS devices.

Also, I couldn't figure out whether it's practicable to limit the footprint of each SVG partition, via `viewBox` settings or something else, so that it doesn't span the entire image extent, but given that the holes rectangles are only present on its particular bit of the roll, and the rest is just blank space, I assume that this shouldn't really affect the shift/redraw performance.

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -125,6 +125,7 @@
   let ppi;
   let svgPartitions;
   let visibleSvgs = [];
+  let entireViewportRectangle;
 
   const createMark = (hole) => {
     const {
@@ -220,6 +221,13 @@
   const partitionHolesOverlaySvgs = () => {
     if (!holeData) return;
 
+    entireViewportRectangle = viewport.imageToViewportRectangle(
+      0,
+      0,
+      imageWidth,
+      imageLength,
+    );
+
     svgPartitions = new IntervalTree();
 
     // Calculate the maximum image length visible at maximum zoom (out)
@@ -276,12 +284,6 @@
     svgs.forEach((svg) => {
       if (visibleSvgs.includes(svg)) return;
       visibleSvgs.push(svg);
-      const entireViewportRectangle = viewport.imageToViewportRectangle(
-        0,
-        0,
-        imageWidth,
-        imageLength,
-      );
       viewport.viewer.addOverlay(svg, entireViewportRectangle);
     });
   };
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -253,11 +253,11 @@
       const holes = holesByTickInterval
         .search(firstTick, lastTick)
         // eslint-disable-next-line no-loop-func
-        .filter(({ y: offsetY, h: height }) => {
-          const yCoord = $scrollDownwards
-            ? offsetY
-            : imageLength - offsetY - height;
-          return yCoord >= firstPixelRow && yCoord < lastPixelRow;
+        .filter(({ y: offsetY }) => {
+          const yCoord = $scrollDownwards ? offsetY : imageLength - offsetY;
+          return $scrollDownwards
+            ? yCoord >= firstPixelRow && yCoord < lastPixelRow
+            : yCoord > firstPixelRow && yCoord <= lastPixelRow;
         });
       if (holes) {
         const svg = createHolesOverlaySvg(holes);
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -167,7 +167,7 @@
     return mark;
   };
 
-  const createHolesOverlaySvg = (firstPixelRow, lastPixelRow) => {
+  const createHolesOverlaySvg = (holes) => {
     const svg = document.createElementNS("http://www.w3.org/2000/svg", "svg");
     const g = document.createElementNS("http://www.w3.org/2000/svg", "g");
 
@@ -178,7 +178,7 @@
     svg.setAttribute("viewBox", `0 0 ${imageWidth} ${imageLength}`);
     svg.appendChild(g);
 
-    holeData.forEach((hole) => {
+    holes.forEach((hole) => {
       const {
         x: offsetX,
         y: offsetY,
@@ -188,21 +188,18 @@
         type: holeType,
       } = hole;
 
-      const yCoord = $scrollDownwards
-        ? offsetY - padding
-        : imageLength - offsetY - height - padding;
-
-      if (yCoord < firstPixelRow || yCoord > lastPixelRow) {
-        return;
-      }
-
       const rect = document.createElementNS(
         "http://www.w3.org/2000/svg",
         "rect",
       );
 
       rect.setAttribute("x", offsetX - padding);
-      rect.setAttribute("y", yCoord);
+      rect.setAttribute(
+        "y",
+        $scrollDownwards
+          ? offsetY - padding
+          : imageLength - offsetY - height - padding,
+      );
       rect.setAttribute("width", width + padding * 2);
       rect.setAttribute("height", height + padding * 2);
       rect.setAttribute("rx", 10);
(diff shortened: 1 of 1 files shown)


src/components/RollViewer.svelte
```diff
@@ -218,7 +217,70 @@
...
+
+    // Calculate the maximum image length visible at maximum zoom (out)
+    const maxViewportImageLength =
+      viewport.viewportToImageRectangle(viewport.getBounds()).height *
+      (maxZoomLevel / viewport.getZoom());
+
+    // Regularize the length of the SVG partitions
+    const partitionLength = Math.ceil(
+      imageLength / Math.ceil(imageLength / maxViewportImageLength),
+    );
```
simonwiles, line 236: Question: what's the logic here?  This seems entirely arbitrary (at first glance, at least) -- I just replaced this with:
```js
const partitionLength = 1000;  // for the sake of argument
```

The logic in `updateVisibleSvgPartitions` will show all and only the needed SVGs.  For the crazy Liszt piece at maximum zoom out on my laptop screen, calculating the `partitionLength` like this results in 3 or 4 SVGS on screen at any given time.  `partitionLength = 1000` results in 31 SVGs on the screen at any given time, and seems to perform slightly better even in the (ad-hoc) stress tests I've done.

How to choose this number?  Is there an inflection point?

src/components/RollViewer.svelte
```diff
@@ -218,7 +217,70 @@
...
+      if (visibleSvgs.includes(svg)) {
+        return;
+      }
+      visibleSvgs.push(svg);
+      const entireViewportRectangle = viewport.imageToViewportRectangle(
+        0,
+        0,
+        imageWidth,
+        imageLength,
+      );
```
simonwiles, line 281: Given that this is running many times per second while the roll is playing, I think this should be moved out of the loop (in fact, it should a component-level variable, not even just a function-level one).

src/components/RollViewer.svelte
broadwell: It's based on intuition that isn't backed up by any practical testing, so from that perspective it's arbitrary, I suppose. 

The intention was to set the partition length roughly equal to the extent of the roll visible at maximum zoom, and then if necessary shrink the partition length until it's an integer multiple of the entire roll length, so there wouldn't be an oddly sized partition at the end. Regardless of whether the latter adjustment is made (its only real motivation is my dislike of remainders), my intuition was a) that this would result in at most 2 SVG partitions being visible at any time, and b) that shifting a small number of SVG partitions that are somewhat large (but still much smaller than the partition that covers the entire roll in the current implementation) might produce better performance that shifting a larger number of smaller partitions.

I'm not sure how intuition point a) can have gone wrong such that 3 or 4 SVG partitions actually are visible at max zoom, but in any case, point b) is a complete guess, and it's quite possible that setting a much smaller partition length will produce better results. I'll give it a try on my system.

src/components/RollViewer.svelte
broadwell: Oops, yeah that definitely shouldn't be in the main loop.

src/components/RollViewer.svelte
simonwiles:
```js

    const rect = document.createElementNS("http://www.w3.org/2000/svg", "rect");
    rect.setAttribute("x", 0);
    rect.setAttribute("y", firstPixelRow);
    rect.setAttribute("width", "100%");
    rect.setAttribute("height", lastPixelRow - firstPixelRow);
    rect.setAttribute(
      "fill",
      `rgba(${Math.floor(Math.random() * 255)},${Math.floor(
        Math.random() * 255,
      )},${Math.floor(Math.random() * 255)},.4)`,
    );
    g.appendChild(rect);
```
in `createHolesOverlaySvg` before the `holeData,forEach` loop makes it fun to watch :)

src/components/RollViewer.svelte
broadwell: Maybe it makes the most sense to set the variable at the beginning of `partitionHolesOverlaySvgs()`?

simonwiles:
It's so great that there's a way to mitigate at least some of the performance issues.  Thanks so much for working on this.

>  I couldn't figure out whether it's practicable to limit the footprint of each SVG partition, via viewBox settings or something else, so that it doesn't span the entire image extent, but given that the holes rectangles are only present on its particular bit of the roll, and the rest is just blank space, I assume that this shouldn't really affect the shift/redraw performance.

Yeah I suspect I might have started with this approach and expended some unnecessary effort.  I'm not sure I would have expected that it would make more-or-less no difference, but it seems that is indeed the case, so there's nothing here that needs "fixing" imo.

Aside from a couple of specific comments I've already made an a couple of minor linting fixes, my main "itch" with this is looping over the entire `holeData` array for every SVG when we already have the holes in an `IntervalTree` (albeit organized by tick).  Would you entertain a commit to generate the SVGs that way (and perhaps only cover the range of the image that actually contains holes, so that we don't have any entirely/mostly empty ones)?

I'm really pleased this works so well -- I was not optimistic about this approach, tbh, and tried many more complicated things without anything like this success.  Thanks for taking it on!

src/components/RollViewer.svelte
simonwiles: I was actually thinking of a component-level variable set in the component's `onMount` -- it's a one-off thing per component mount.

broadwell: Yeah, in the current PR, `partitionHolesOverlaySvgs()` is called as a result of `openSeadragon.addOnceHandler("update-viewport" ...`, so setting the variable in that function seemed like a good fit logically and practically, because it appears that the full extent of the `viewport` isn't really known until after the first `update-viewport` event.

src/components/RollViewer.svelte
broadwell: Now that I think about it more, my intuition was definitely off; obviously the number of individual hole SVGs being shifted at any given time is what correlates with the performance issues, not the size or number of the enclosing parent SVG partitions being shifted. So the optimization goal should be to size the partitions so as to minimize the number of not-on-screen SVGs that are being shifted at any time (or zoom level), while also avoiding the overhead of exhaustively checking whether a given hole is on screen at that time before displaying its SVG -- hence the use of the `svgPartitions` intervalTree. So a smallish `partitionLength` is almost certainly preferable, though I have no idea where the inflection point actually lies. 1000 seems plausible, or else just use the length of image visible at zoom=1?

broadwell: > in createHolesOverlaySvg before the holeData,forEach loop makes it fun to watch :)

Cool -- it also shows where my intuition a) went awry: in the PR, the `partitionLength` is computed based on the amount of the roll image visible at max zoom *when it first loads* -- that is, with the center point at tick=0; the issue is that the roll image doesn't fill the entire viewport when zoomed out at tick=0, so a `partitionLength` is chosen that results in more than 2 partitions being visible when the entire roll eventually pans into view as it plays.

Anyway it's kind of a moot point given our discussion above.

Screen Shot 2021-09-24 at 1 01 35 PM

broadwell:
> Would you entertain a commit to generate the SVGs that way (and perhaps only cover the range of the image that actually contains holes, so that we don't have any entirely/mostly empty ones)?

Sure -- it's definitely doing a lot more work than strictly needed. Would you like to give it a shot?

src/components/RollViewer.svelte
simonwiles: > minimize the number of not-on-screen SVGs

You mean "minimize the number of not-on-screen hole `<rect/>`s, of course, but yeah, this seems right (again, not exactly what I would have expected working from first principles -- but then I don't see why doing *more work* in javascript is resulting in better performance here; we've obviously hit something that is not well optimized in browsers (and it's clear that Chrome is worse that Firefox here, too)).  The query time for the interval tree should scale as `O(log n)`, so I wouldn't worry about the impact of adding more SVGs (actually, even for the likely upper bound on a long roll with `partitionLength < 1000` it seems likely that an interval tree could be outperformed by simpler structures, but that would be `√all evil` at this point :)).

src/components/RollViewer.svelte
```diff
@@ -218,7 +216,70 @@
...
+    const svgs = svgPartitions.search(firstImagePixel, lastImagePixel);
+
+    // Remove any currently displayed SVG overlays that don't overlap with the
+    // viewer window
+    visibleSvgs.forEach((visibleSvg) => {
+      if (!svgs.includes(visibleSvg)) {
+        visibleSvgs.splice(visibleSvgs.indexOf(visibleSvg), 1);
+        viewport.viewer.removeOverlay(visibleSvg);
+      }
+    });
```
simonwiles, line 267: Mind if I update this so that `visibleSvgs` is treated as immutable (similar to the treatment of `marks` in `highlightHoles`)?

broadwell: Sure -- that seems fine to me.

PR #105, opened 2021-09-17
Fix weird Safari RollSelector searchbar bug

broadwell:
When using Safari on Mac OS or iOS, there's an odd bug whereby entering text in the roll selector searchbar and then deleting all of it somehow prompts Safari to leave an invisible `<br>` tag in the input field, which then results in all of the roll names with "br" in them being highlighted until the selector is closed and reopened.

This is one possible fix; there may be better ones.

Screen Shot 2021-09-16 at 11 02 25 PM

### DIFF
--- a/src/ui-components/FilteredSelect.svelte
+++ b/src/ui-components/FilteredSelect.svelte
@@ -252,7 +252,7 @@
     if (!input.innerHTML) return;
 
     const filteredText = normalizeText(
-      input.innerHTML.replace(/[&/\\#,+()$~%.'":*?<>{}]|nbsp;/g, " "),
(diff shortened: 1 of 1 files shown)


broadwell:
Of course it's not clear that the app really can get sufficient performance out of Safari such that the user experience when playing rolls is acceptable. My testing so far has indicated that neither mobile nor desktop Safari can provide both a smooth-scrolling OSD and synchronized audio playback, even on fairly high-end devices. But Safari (iOS or Mac OS) works fine in an "audio-only" configuration (no OSD viewer, but smooth MIDI playback), so this PR is made with that potentiality in mind.

simonwiles (approved):
Yeah, this is an acceptable solution to me.  Thanks!

PR #110, opened 2021-09-29
Some UI Updates

simonwiles:
This is a mixed collection of updates to the UI, and a certain amount of component refactoring and DRYing up.  I have a few more things I've been working on that might have wound up in here, but this is the kind of branch that could just go on and on, so I figured best to PR it now and start a new branch :)

Highlights include:

* A new typeface  -- the biggest motivation here is that `serif` and `sans-serif` will result in different typefaces depending on the browser/platform combo, and so the use of space isn't consistent.  They were always suppose to be placeholders anyway.  I think this typeface looks okay -- of course it can be up for discussion.  We also might consider self-hosting the font files.
* Better arrangement of the `<RollDetails/>` component -- there's more could be done here, but this has been sitting in a branch for literally months :)
* Nice cross-browser checkboxes
* Styling of the `::selection` pseudo-element
* Reworking of the slider controls -- this includes restyling of the sliders themselves, and also encapsulating them with their associated labels.  This makes the code a little cleaner, but also allows for a couple of UX improvements and a central place to implement them.  As of this PR they have better mousewheel and click behaviours, but there are a couple of other things I have in mind that will be more approachable now, if they seem like good ideas.
* A new "fit-width" icon (as per George's suggestion).
* Other minor styling and code updates.

Some of these are recent, others have been cherry-picked from branches I started months and months ago.


broadwell (commented):
This looks amazing! The skinny scrollbars might need a little getting used to, but they're definitely an improvement, as is the consistent mousewheel behavior of the scrollbars (even though I'm still not sure it's really needed).

I'm definitely in favor of self-hosting the font files eventually.

Just one request, which could go in a later PR: can we add a `word-wrap: break-word` where appropriate, because Chrome (on Mac OS, at least) refuses to break the URLs?
Screen Shot 2021-09-29 at 4 46 06 PM

simonwiles:
> Just one request, which could go in a later PR: can we add a word-wrap: break-word where appropriate, because Chrome (on Mac OS, at least) refuses to break the URLs?

Yeah, okay, I'm down with that :)  There is horizontal overflow there, but I guess Mac OS is hiding the scroll bars so it's not really discoverable?

`word-break: word-break;` is deprecated, fwiw, so I'll apply `overflow-wrap: anywhere;` to the same effect (`overflow-wrap: word-break;` might arguably be marginally better, but it doesn't have quite as wide support -- it's more-or-less a wash anyway).

broadwell:
Thanks. I guess Firefox has some built-in logic to break URLs on slashes even when horizontal overflow is available? Interesting. Anyway, I'd definitely prefer for the PURL link eventually to be shortened/buttonized so that we don't need to use panel space to spell out the full `https://purl.stanford.edu` bit. Not much information gain there. Maybe this will happen in that nifty metadata buttons branch you previewed?

Also, I think we can drop the `Call No` field for good. Shelfmark/callno data only seems to be available for a subset of the current records, and likely won't be available at all in the new MODS format. But that can wait until a later PR.

simonwiles:
Yep, I dropped the "Call No" and replaced the "PURL" with a button/pill labelled "SDR" (as it is done on the SUPRA site) in that branch -- still unclear how I can move forward with that branch, though, tbh.  Anyway, thanks for merging this, I'll get it deployed.

PR #111, opened 2021-10-02
Add persistent velocity curve editors for top, middle, bottom thirds of keyboard

broadwell:
This branch supersedes [`velocity-spline-curve`](https://github.com/sul-cidr/pianolatron/tree/velocity-spline-curve).

It made the most sense to store the interpolated spline curves as well as the user-provided anchor points persistently so that the curves' effect on the keyboard input velocities and the configurations of the editors would be retained across sessions, and importantly the former would take effect immediately upon starting the app, without requiring a visit to the Audio Settings tab.

The editor widgets in the sidebar (there are now three of them) are more responsive and the styling is a bit better, though both aspects could use further refinement.


broadwell:
OK, after merging in #113 I think this is much improved and ready to merge, but I'll leave the final call up to you.

simonwiles (approved):

PR #112, opened 2021-10-04
Send and receive note and pedal events to/from connected devices via WebMIDI

broadwell:
The Web MIDI functionality, when available (currently only on Chrome/Edge including Android, though for some others there's this sketchy [polyfill/shim](https://github.com/jazz-soft/JZZ)) is pretty tightly coupled to the `SamplePlayer`'s operations, but it's tidier to have a distinct `WebMidi` component that can persist for the lifetime of the app.*

It seemed preferable to avoid needing an actual MIDI connections config menu, so to this end the component connects automatically to all available MIDI devices, listening to and sending messages to all of them. It thus was necessary to complicate the note and pedal on/off logic slightly so that the app would be a good citizen of the MIDI device network and not trigger any MIDI through feedback loops. The component also makes what I think is pretty successful use of the nifty `Notification` functionality to keep the user updated about MIDI device comings and goings (and also the overall availability of Web MIDI or lack thereof from the browser) in a fairly unobtrusive way.

Note that this functionality can be tested without needing any special MIDI hardware; on Mac OS, the combination of the Inter-Application Communication (IAC) driver and MidiKeys works quite well. [This page](https://www.skratchdot.com/2016/01/creating-virtual-midi-ports-on-osx/) lists similar combos for other OSes. 

\* It also would be quite straightforward for this component to make a downloadable MIDI "recording" of the note and pedal events sent to it, even if the browser doesn't support actual Web MIDI device interactions.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -287,8 +287,8 @@
         {/if}
       {/if}
     </FlexCollapsible>
-    {#if appReady}
-      <div id="roll">
+    <div id="roll">
+      {#if appReady}
         <RollViewer
           bind:this={rollViewer}
           imageUrl={currentRoll.image_url}
@@ -296,16 +296,16 @@
           {holesByTickInterval}
           {skipToTick}
         />
-        {#if $userSettings.showKeyboard && $userSettings.overlayKeyboard}
-          <div id="keyboard-overlay" transition:fade>
-            <Keyboard keyCount="88" {startNote} {stopNote} />
-          </div>
-        {/if}
-      </div>
-      <FlexCollapsible id="right-sidebar" width="20vw" position="left">
-        <TabbedPanel {playPauseApp} {stopApp} {skipToPercentage} />
-      </FlexCollapsible>
-    {/if}
+      {/if}
+      {#if $userSettings.showKeyboard && $userSettings.overlayKeyboard}
+        <div id="keyboard-overlay" transition:fade>
+          <Keyboard keyCount="88" {startNote} {stopNote} />
+        </div>
+      {/if}
+    </div>
+    <FlexCollapsible id="right-sidebar" width="20vw" position="left">
+      <TabbedPanel {playPauseApp} {stopApp} {skipToPercentage} />
+    </FlexCollapsible>
   </div>
   {#if $userSettings.showKeyboard && !$userSettings.overlayKeyboard}
     <div id="keyboard-container" transition:slide>
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -178,6 +178,41 @@
     holesByTickInterval = new IntervalTree();
   };
 
+  const initWebMidi = () => {
+    if (navigator.requestMIDIAccess) {
+      if (!$userSettings.midiMessageSeen && !$showWelcomeScreen) {
+        notify({
+          title: "MIDI in/out available",
+          message:
+            "Connect a digital piano or other MIDI divice to send/receive keyboard events.",
+          closable: true,
+        });
+        $userSettings.midiMessageSeen = true;
+      }
+      navigator.requestMIDIAccess().then((midi) => {
+        midi.onstatechange = (e) => {
+          // Print information about the (dis)connected MIDI controller
+          notify({
+            title: "MIDI device change",
+            message: `${e.port.name} ${e.port.manufacturer} ${e.port.state}`,
+            timeout: 4000,
+            closable: true,
+          });
+        };
+      });
+    } else {
+      if (!$userSettings.midiMessageSeen && !$showWelcomeScreen) {
+        notify({
+          title: "MIDI in/out not available",
+          message:
+            "This browser does not support connecting to a digital piano or other MIDI device.",
+          closable: true,
+        });
+        $userSettings.midiMessageSeen = true;
+      }
+    }
+  };
+
   const loadRoll = (roll) => {
     mididataReady = fetch(`./midi/${roll.druid}.mid`)
       .then((mididataResponse) => {
@@ -221,6 +256,7 @@
           url.searchParams.set("druid", currentRoll.druid);
           window.history.pushState({ roll: currentRoll }, "", url);
         }
+        initWebMidi();
       },
     );
   };
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -83,6 +83,7 @@
   } from "./ui-components/Notification.svelte";
   import FlexCollapsible from "./ui-components/FlexCollapsible.svelte";
   import LoadingSpinner from "./ui-components/LoadingSpinner.svelte";
+  import WebMidi from "./components/WebMidi.svelte";
 
   import catalog from "./config/catalog.json";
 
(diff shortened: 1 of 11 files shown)


simonwiles:
This is great!

I've had a fun evening trying to set up virtual MIDI devices on my Arch system, and playing around with this.  I confess I'm a bit out of my depth with MIDI sequencers and synthesizers and DAWs and so on, but I've been able to record some output from the Pianolatron (and transcribe it to musical notation!) using [Rosegarden](https://www.rosegardenmusic.com/), and then play it back from a MIDI device into the Pianolatron, adjusting the audio options on-the-fly (but not the tempo, which I suppose makes sense).  I do have a keyboard here with MIDI capabilities, but that's in my eldest's room and he's asleep, so I might give that a whirl over the weekend :)

I have some thoughts about the functionality:
* I wholeheartedly support the idea of connecting to all available MIDI devices automatically -- for someone new to that side of things it all "just worked"™, and I assume those who have complicated set ups also know how to configure things correctly in their audio subsystem (I've been down the rabbit-hole far enough tonight to have been playing with [a new patch bay written in Rust](https://gitlab.freedesktop.org/ryuukyu/helvum) :)).  I've certainly no desire to get into writing UI for that (the new keyboard shortcuts updates were gnarly in the end), but I wonder if we mightn't hide all the web-midi stuff behind a "Enable/Disable Web-MIDI" button, or something like that?  While the notifications are mostly unobtrusive, the bar for "too obtrusive" is really low for functionality most folks won't ever use, imo.  The `$midiMessageSeen` goes some way to addressing this (especially in non-WebMIDI-supported environments), but I still get multiple notifications on every page load in Chromium.
* I'm a bit unclear on what the value of having the Pianolatron accept MIDI input really is.  In addition to making the code a little more complicated (although you've done a nice job of keeping that clean), I wonder if it also makes some of the behaviour a bit awkward and ambiguous.  I can play MIDI back through the Pianolatron (or from a physical keyboard, I suppose) and disable pedaling and expressions...  What does that even mean?  Can you sell me on the value of MIDI input?

w/r/t the code, I'm very happy with the organization.  My only comment would be that the `<WebMidi/>` component could perhaps be a direct child of the `<SamplePlayer/>` component, since the former's functionality is entirely scoped beneath the latter.  The code is good -- there are some linting errors I've cleaned up in `midi-in-out-updates`, and some stylistic things I'm tempted to suggest some changes to.  I did make it crash once, but I need to be more systematic to see if I can reproduce that reliably.  I'm happy to work on the notification-related issues you mentioned in d2d83de8.

I'll update if I can reproduce the crash, but I wanted to get this down now.

broadwell:
> I wonder if we mightn't hide all the web-midi stuff behind a "Enable/Disable Web-MIDI" button, or something like that?

Yeah, this is probably warranted. The notifications definitely can get rather chatty, and reducing them to a level that most users would consider truly unobtrusive seems like it would difficult to achieve and probably wouldn't be welcomed by the small % of users who actually want to take advantage of the Web MIDI features. So an enable/disable config button (default: disabled) is probably the best option.  
Re: the notifications, it's a bit odd that on my system, there's a `midi.onstatechange`-triggered notification about the inter-app MIDI system being available when the app starts **and** when the first note is played in the app; this seems rather redundant and maybe one of them could be quashed, but otherwise the messages that pop up seem appropriately informative.

> I'm a bit unclear on what the value of having the Pianolatron accept MIDI input really is.

Yes this is definitely borderline usefulness-wise, at the moment. As long as the user can both listen to and play the connected HW or SW MIDI keyboard devices that are receiving the Pianolatron's MIDI output, then this should cover the two main use cases the PIs have mentioned for which combining input from the app and a separate device would be useful, specifically 1) applying partial sustain pedaling to MIDI events coming from the Pianolatron, and 2) playing a duet on a separate keyboard with the music being produced by the app. In the case of 2), it might be somewhat desirable to be able to see the keys and pedals of the Pianolatron piano app responding to both the roll's and the human's input, but that's a pretty limited benefit.  

One situation in which accepting MIDI input in the app would be called for would be if the app provides the ability to **record** a MIDI performance, including any input that a user might provide via an external keyboard. I have a demo of this partially working via a customized version of [midi-writer-js](https://github.com/grimmdude/MidiWriterJS), but it's not clear how much more effort is needed to get it to 100% functionality, and I'm also not sure whether this is really desirable, given that many external MIDI devices also provide the ability to record MIDI performances, including any inputs they'd receive from the Pianolatron app.

simonwiles (approved):

simonwiles:
Okay, this is now ready for your perusal and, hopefully, approval.  I think this is probably how it should be done, tbh.  It's complete and neat as it stands, but leaves space to add more fine-grained control and additional UI should it be desired.

broadwell:
The MIDI panel looks great and is, I agree, the best way to handle the UI/notification features.

I've noticed one bug so far, which is that if I have the MIDI panel open with WebMidi enabled (Chrom* only, obvs) and then select a different roll, the roll viewer goes blank and doesn't display any tiles or overlays, even though the new roll still plays. Also, the right-hand panel doesn't switch back to the basic settings, as it does when loading a different roll in any other scenario and with any other settings panel selected. So maybe this is an issue with how the MIDI settings panel is destroyed (or not) in this scenario?

Also, I did once manage to get the player into a state of "runaway pedal" (when the sustain pedal goes haywire, flipping on/off continuously), which I've occasionally seen before with previous versions of the branch. But this only happened when I accidentally disconnected and reconnected the network cable during a roll playback on a connected MIDI piano, which caused degraded performance from the app all around. So it's probably just worth keeping an eye out for it.

simonwiles:
> I've noticed one bug so far, which is that if I have the MIDI panel open with WebMidi enabled (Chrom* only, obvs) and then select a different roll, the roll viewer goes blank and doesn't display any tiles or overlays, even though the new roll still plays. Also, the right-hand panel doesn't switch back to the basic settings, as it does when loading a different roll in any other scenario and with any other settings panel selected. So maybe this is an issue with how the MIDI settings panel is destroyed (or not) in this scenario?

Yeah, this is caused by sveltejs/svelte#5268 again, as you guessed.  Removing the transitions on the MIDI device "items" (the `<li/>s` in the "Connected..." lists) solves the problem, and just using intro transitions and removing the outro transitions is probably sufficient.  However, in this case I've just committed a better fix, which is to not recycle that part of the component tree at all when the roll is changed.  This is a change that is bigger in scope, though, and could potentially affect other things, so I'd be grateful if you could give it another test and help me be confident it hasn't broken something elsewhere.  I'm about 95% confident we're good... :s

simonwiles:
> Also, I did once manage to get the player into a state of "runaway pedal" (when the sustain pedal goes haywire, flipping on/off continuously), which I've occasionally seen before with previous versions of the branch.

I've noticed this for this first time recently too, but I can't work out how to reproduce it reliably yet.  Presumably it's a feedback loop between the `<WebMidi/>` component and the `<SamplePlayer/>` component.  One thought I do have is that it adds weight to the case for disabling WebMIDI by default even on MIDI-enabled browsers (I've seen this even without an external device connected) -- right now I've defaulted to on for Chromium-based browsers.  What do you think?

PR #113, opened 2021-10-07
Velocity spline curves updates

simonwiles:
Here is my reworking of the `velocity-spline-curves` branch.  The major refactoring is the creation of a `<VelocitySpliner/>` component which encapsulates the functionality, and (pretty much necessitated by that) the splitting out of the stores for the low-, mid-, and high-range curves (I'm still not clear in my own mind how Svelte stores are best structured, but in general it seems like smaller is nearly always working out better).

I've also taken the liberty of updating the styling and UI a little -- the theming and the reset button are improved; the final two commits are perhaps a little more debatable.  Let me know what you think :)

Edit: if/when you're happy with this you can merge it (it will go into your branch), and then I'll merge #111.  Alternatively, you could change the base of this PR (using the edit button above on the GitHub UI) to `main` and merge it directly -- it will take the contents of both branches.

### DIFF
--- a/src/components/VelocitySpliner.svelte
+++ b/src/components/VelocitySpliner.svelte
@@ -141,7 +141,7 @@
     <div bind:this={splinerContainer} class="spliner-container" />
     {#if $keyboardRegion.velocityPoints}
       <button
-        transition:fade
+        in:fade
         use:tooltip={"Reset"}
         on:click={() => {
           keyboardRegion.reset();


broadwell (changes requested):
The spliner code is much cleaner now, especially after moving to a dedicated component. The reset button is a great improvement to the appearance of the spliners, as is their ability to conform to the selected color scheme! And the axis labels look at least as good as anything I could think of. Also, I'm not sure why I initially selected a smaller point radius, but 7 seems like a suitable value.

There's just one bug that appears to have cropped up, assuming I've merged and rebuilt everything correctly (the proof will be in the reproducing): I'm not sure of the root cause, but now after selecting the audio settings menu on the right panel (the piano icon), then trying to load a new roll, the roll viewer doesn't load the image of the new roll for me, at least on Firefox and Chrome on MacOS. A clue: previously, the right panel would switch from the piano panel back to the basic settings panel (the gear icon) on roll change (which is maybe not the ideal behavior, but at least the new roll image would load at that point), and with this update it stays on the piano settings panel while neglecting to load the new roll image. Ideas about how to fix this, if it's indeed legit?

Also, I really need to figure out why my linter apparently has gone on strike.

PR #114, opened 2021-10-11
Add configurable delta values for keyboard controls

simonwiles:
This PR drops the <kbd>Ctrl</kbd> and <kbd>Shift</kbd> modifiers for volume and tempo key controls, and adds a single configurable "augment" key command for bass volume and tempo ("left hand" -- defaults to <kbd>w</kbd>) and volume and treble volume ("right-hand" -- defaults to <kbd>[</kbd>).

To make the increment values (`delta`, `augmentedDelta`) configurable, slider controls have been added to the "Keyboard Controls" dialog, and since that has become a little crowded now, it's been split into a tabbed arrangement.  I hope you'll agree it's a passable solution to an awkward problem.  A few bits and pieces of the stores, controls, and config had to be shuffled to make this work nicely and it took a fair bit of iterating to get it all working properly, but I think it's ready for destruction testing now :)

### DIFF
--- a/src/components/KeyboardShortcutEditor.svelte
+++ b/src/components/KeyboardShortcutEditor.svelte
@@ -54,29 +54,38 @@
   }
 
   p.error-message {
-    border-radius: 5px;
-    border: 1px solid red;
     color: red;
     padding: 0.25em;
+    margin: 0.5em 0 0;
+  }
+
+  .panels {
+    display: flex;
+    overflow-x: hidden;
+    overflow-y: auto;
+    padding-top: 4px;
+    position: relative;
   }
 
   dl {
     display: grid;
+    flex: none;
     gap: 0 0.25em;
     grid-auto-rows: min-content;
     grid-template-columns: auto auto;
-    height: 400px;
     justify-content: space-between;
-    overflow-x: hidden;
-    overflow-y: auto;
-    padding-top: 4px;
+    width: 100%;
+
+    &:not(:first-child) {
+      margin-left: -100%;
+    }
   }
 
   p.reset {
     display: flex;
     align-items: center;
     justify-content: flex-end;
-    margin: 1em 0;
+    margin: 0.5em 0 0;
   }
 
   button {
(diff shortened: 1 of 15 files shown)


src/config/keyboard-shortcut-config.js
```diff
@@ -122,4 +129,63 @@ export const alternativeIndicatorText = {
   ArrowDown: "↓",
   ArrowLeft: "←",
   ArrowRight: "→",
+  ControlLeft: "L_Ctrl",
+  ControlRight: "R_Ctrl",
+  ShiftLeft: "L_Shift",
+  ShiftRight: "R_Shift",
+  AltLeft: "L_Alt",
+  AltRight: "R_Alt",
+  MetaLeft: "L_Meta",
```
broadwell, line 138: So it turns out that on Mac OS, the `code` attributes of the <kbd>command</kbd> keys actually are `OSLeft` and `OSRight`, while the `key` value for both is just `Meta`. As predicted, this is getting into the weeds pretty quickly, but maybe we can just add a couple more mappings here, specifically
`OSLeft: "L_Meta",`
`OSRight: "R_Meta",`
and then call it good enough.

src/components/KeyboardShortcutEditor.svelte
```diff
@@ -79,18 +143,77 @@
...
+        "VOLUME_DELTA",
+        "VOLUME_AUGMENTED_DELTA",
+        "---",
+        "TREBLE_VOLUME_UP",
+        "TREBLE_VOLUME_DOWN",
+        "TREBLE_VOLUME_DELTA",
+        "TREBLE_VOLUME_AUGMENTED_DELTA",
+        "---",
+        "RIGHT_HAND_AUGMENT",
+        "---",
```
broadwell, line 191: This is good, although now that I'm looking at it, I can't help but think that rather than two `<hr>`s, a dark line that goes all the way across the panel, as with the bottom border of the tab groups, might be even better at highlighting the separation of the "left" and "right" halves. But if that's at all difficult to accomplish here, then I'm fine with leaving this as is.

src/config/keyboard-shortcut-config.js
simonwiles: Yeah, that's the assumption...  What's the old adage about making an ass out of u and me? :D

PR #116, opened 2021-12-16
Upgrade OpenSeadragon to v3.0.0

broadwell:
This doesn't seem to cause any problems, at least on Firefox and Chrome for MacOS. It's possible that the update brings some performance improvements, but if so, they're subtle.
This could be folded into a PR with upgrades of other components, if applicable, but is also OK to go in on its own.

The version release notes:
https://github.com/openseadragon/openseadragon/releases/tag/v3.0.0


simonwiles (approved):
Yeah, this all checks out.  Might be able to simplify some of the reimplementation of event handlers that happens in our codebase now.

PR #118, opened 2022-01-31
[WIP] MIDI recorder using MidiWriterJS

broadwell:
OK, this works and is ready for discussion but not merging yet --  hence the messy commit log.

It relies on the customized version of MidiWriterJS [here](https://github.com/broadwell/MidiWriterJS/tree/web-piano-with-control-timings), which probably would need to be inlined or something. The current upstream of MidiWriterJS is unable to generate the proper timings when interleaving multiple pedal (controller) events with note events; most of the changes in the downstream version involved essentially cloning the existing timing/serialization code for note down/up events and applying it to pedal down/up events. The functionality now being provided by the customized MidiWriterJS was _just_ complicated enough that it seemed preferable to adapt the existing code rather than to implement all of it from scratch.

On the Pianolatron side, there's a pretty good case to be made that the MIDI recorder functionality should be separated from the `WebMidi` component, or maybe it's fine to commingle them (it's certainly quite convenient to tie the recording logic into `sendMidiMsg()`) and the component just needs a different, more all-encompassing name. There are also probably several other improvement points lurking in the code, but hopefully it will be productive to talk about this now.


broadwell:
Closing this in favor of the in-progress implementation in [midi-writer](https://github.com/sul-cidr/pianolatron/tree/midi-recording), which uses the MIDI output capabilities of [ToneJS/MIDI](https://github.com/Tonejs/Midi). Its features for generating MIDI and particularly for incorporating controller events (such as pedaling) into the output MIDI seem much more mature than those in MidiWriterJS; using the latter would require forking our own branch, while ToneJS/MIDI can be used pretty much out of the box.

PR #119, opened 2022-02-17
Misc. Updates

simonwiles:
This is a bunch of tiny commits that have been hanging around for ages.  Please merge whenever it's convenient.  Cheers!


broadwell:
This all looks good to me. I recently began seeing the `caniuse` complaints, so I'm glad that's going away.

PR #120, opened 2022-03-04
Refresh roll data, accommodate new types in metadata search/display

broadwell:
These updates bring the playable rolls into sync with the updated parsing code at [https://github.com/sul-cidr/roll-image-parser](https://github.com/sul-cidr/roll-image-parser) (plus a [pending update](https://github.com/sul-cidr/roll-image-parser/tree/88-note-continuations-fix) to improve hole segmentation for the 88-note rolls), adding a small number each of green Welte, Licensee (De Luxe) and Duo Art rolls to the set.

Some minor code additions were necessary, including adjustments to the metadata catalog fields and the formatting regexp for the roll searchbar.

Adding the new roll types, which all use continuous control holes rather than the "lock-and-cancel" format of the Welte red (T-100) rolls, exposed an issue with the hole overlay visualizations in which the identifying flags of long holes aren't visible when moused over/activated if they're long enough that the top of the hole extends past the top of the view window. This can be addressed in a future PR.


simonwiles:
Yep, this all looks grand, thanks.

* The duo-art rolls are really quite lovely to look at (perhaps it's just that they're a change from all the welte ones!)
* The speed at which the images from the SDR is, indeed, a little frustrating
* The Chopin welte-green one plays some interesting visual artifacts from the leader, but I expect you've already noticed that!

Merging now, and I'll do a deploy shortly afterwards.

simonwiles (approved):

PR #121, opened 2022-04-15
Correct red Welte expression MIDI files

broadwell:
A stealth bug in the `midi2exp` code, described in the https://github.com/pianoroll/midi2exp/pull/3 PR, has caused expression MIDI files for red Welte rolls generated since July 2021 to lack proper soft pedal events. This PR includes the corrected versions of these files (and their accompanying JSON metadata files).


src/config/catalog.json
```diff
@@ -1173,7 +1173,7 @@
     "number": "3398",
     "publisher": "Welte-Mignon",
     "title": "Delibes-Dohnányi/Blumen - \"naïla\": Walzer",
-    "type": "welte-red"
+    "type": "88-note"
```
simonwiles, line 1176: What's going on with this and the other three which have been reassigned from `welte-red` to `88-note`?  They all look like `welte-red` rolls to me?  Am I missing something?  (sorry!)

broadwell: This is a good catch -- most likely the MODS files for those records have been modified so that the methods previously used to determine the roll type no longer work (sigh). I'll look into it.

broadwell: So yeah, it was a combination of inconsistent MODS metadata and the changes I made to [pianolatron-data/build-pianolatron-files.py](https://github.com/pianoroll/pianolatron-data/blob/main/build-pianolatron-files.py) to assign the proper roll type to our one 65-note roll. They _should_ all be fixed now.

PR #124, opened 2022-08-18
Framework Updates

simonwiles:
This PR updates the codebase to the stable Astro 1.x APIs.  The biggest difficulty was that this update moves from the (now very) old snowpack-based versions of Astro to the new vite-based ones, and this made the slightly gnarly SASS/SCSS setup difficult to replicate (the astro-svelte-sveltepreprocess-scss integration isn't well documented at this time) -- I very nearly started on ripping out all the SASS, tbh, but in the end found the way to get the pipeline working properly.

This PR also updates all other dependencies to their latest versions **with the exception of `webmidi`**.  WEBMIDI.js seems to have undergone some major changes, some of which are possible breaking changes (see https://webmidijs.org/docs/migration/), and which will require more testing that I'm able to give it right now.  Something to think about if we're touching all the Web MIDI code again in the near future, though.

* `rm -rf node_modules && yarn` required, of course.


broadwell (approved):
It's great that the Pianolatron has finally been brought kicking and screaming up to Astro 1.0.x, and thanks for keeping the rest of the packages current. I do get warnings now about CanvasSpliner being old-school JS, but testing has shown everything to be fully functional.
It looks to me like the upgrades to webmidi shouldn't be too much trouble to integrate, but I agree it's best to handle that in a later update.

broadwell:
I'll just mention that `vite` does spit out a couple of complaints (see below) during `astro build` that might be worth addressing at some point, but certainly aren't urgent.
```
vite v3.0.5 building for production...
transforming (2) src/Pianolatron.svelte12:19:45 PM [vite-plugin-svelte] /Users/peterbroadwell/Documents/piano_rolls/pianolatron/src/components/KeyboardControls.svelte:21:0 Unused CSS selector "button"
19: }
20:
21: button {
    ^
22:   border-radius: 4px 4px 0 0;
23: }</style>
transforming (67) src/components/RollViewerControls.svelte12:19:45 PM [vite-plugin-svelte] /Users/peterbroadwell/Documents/piano_rolls/pianolatron/src/components/VelocitySpliner.svelte:46:0 Unused CSS selector "button[disabled]"
44:   stroke: black;
45: }
46: button[disabled] {
    ^
47:   cursor: default;
48:   opacity: 0.5;
```

simonwiles:
> I do get warnings now about CanvasSpliner being old-school JS, but testing has shown everything to be fully functional.

Actually the issue is that CanvasSpliner is new-school (i.e. ESM) but doesn't explicitly identify itself as such.  It's a simple fix, but the upstream doesn't seems well-maintained and it seems a PR would be unlikely to result in a new release.  I've no desire to fork it, tbh, so it's either live with the warning or vendor it in, I think.  Let's see how irritating we find it over the next week or two :)

PR #125, opened 2022-09-08
Enable tooltips for menu selector, roll navigation and piano show/hide control

broadwell:
I'm not sure if this overlaps with other UI-related branches that are in the works, but anyway I think casual users would appreciate some basic tooltip hints/explanations about what functionality groupings are meant to live under each of the menu icon buttons at top right. The roll navigation buttons and piano pop-out/in/under icon buttons are quite a bit more self explanatory, but I still suspect that mouse-users would be positively inclined towards tooltips on these.

This implementation is meant to coexist with the use of the PopperJS tooltips for the key bindings menu -- that is, not supplying an icon's `alt` property must means that no default tooltip will be shown, clearing the stage for a PopperJS tooltip. It's also somewhat orthogonal to any specifically ARIA-oriented functionality, for better or worse.

### DIFF
--- a/src/components/KeyboardControls.svelte
+++ b/src/components/KeyboardControls.svelte
@@ -43,6 +43,7 @@
     label="Show/Hide Piano Keyboard"
     height="24"
     width="24"
+    alt="Show/Hide Piano Keyboard"
   />
   {#if !outside}
     <IconButton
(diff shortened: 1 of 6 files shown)


broadwell:
Closing, as the desired tooltips/alt text strings are now fully available for these buttons in the `develop` branch.

PR #126, opened 2022-09-09
Expressionizer Classes

simonwiles:
This is a bit of a monster of a PR.  The initial commits come from my original branch (the earliest of which goes back a year!), then I cherry-picked your commits on top of that, then there's a load more commits which are my refactoring of your expressionizer code.

The code for drawing the expression curves *is* in here, because it's part of what I cherry-picked from your branch, but I disabled it while developing, as it makes the rolls basically unplayable for me with in-app expression on (same is true on the staging site).  When you're happy to merge this I'll move on to looking at that.

There is still much more that could be done here, and the code still feels quite "unpolished", but if I've not messed up any refactoring or broken any functionality then this will hopefully be a solid foundation.  Since I'm out tomorrow I wanted to get this up tonight.

I'm hoping you'll be able to sanity-check the end result, and perhaps you could go through the expressionizer code and clean up / add / remove the comments as appropriate and see what, if any, parts need to be worked on.  Lmk if you have questions!  Thanks!

(draft PR for now)


broadwell:
This is great work, and the class hierarchy should provide a very strong foundation for the expression emulation wing of the code. The results of the various expression boxes all look good to me so far. I still need to take a stab at refining the comments in the expressionizer code, but for now, I figured I should flag a couple of instances of problematic behavior I've noticed around the in-app expression components:
- If a roll of a different type is selected after enabling In-App expression, the Expression Settings don't seem to change to reflect the new roll type, and judging from the visualizations and playback oddness that ensues, I'm guessing that the expression boxes aren't being successfully swapped.
- The Reset to Defaults button under the Expression Settings doesn't work -- it just deletes the entire input menu. Relatedly, though, the appearance of the loading animation when an Expression Settings value is changed is a great and much-needed addition to the UI.
I can try to track down the sources of these glitches, but let me know if some quick fixes come to mind.

simonwiles:
> * If a roll of a different type is selected after enabling In-App expression, the Expression Settings don't seem to change to reflect the new roll type, and judging from the visualizations and playback oddness that ensues, I'm guessing that the expression boxes aren't being successfully swapped.
> *The Reset to Defaults button under the Expression Settings doesn't work -- it just deletes the entire input menu.
> ...
> I can try to track down the sources of these glitches, but let me know if some quick fixes come to mind.

Thanks, no worries, I'll take these.  They're both "next-step" issues :)

> Relatedly, though, the appearance of the loading animation when an Expression Settings value is changed is a great and much-needed addition to the UI.

Yeah, this comes for free now, on account of using the way the stores are hooked up.  Some refining may be possible and/or desirable, but not high-priority I think.

Let's keep working on these bits in this branch?

broadwell:
Yes, it makes sense that switching smoothly between roll types is a secondary concern to getting the individual in-app expression modes working, and those have all checked out for me (including the discrepancies you noticed between the expressive MIDI and in-app expression for some of the Welte Licensee rolls, which will need investigation eventually).

Relatedly, another highly desirable next-step feature would be an expression curve overlay display that doesn't sabotage the actual roll playback. Drawing it certainly should be optional (and probably not the default), but it's very helpful for debugging/double-checking what the in-app expression logic is up to. This could happen in a separate PR, though. Otherwise, yes, let's keep plugging away at this branch.

broadwell:
Just noting a couple of other bugs that have come up, which are fairly minor and likely to be fixed in the course of refactoring that is already somewhere on the to-do list:
- Selecting in-app expression for a Duo-Art roll causes a popup error:
```
MIDI Data Error!
Error: `low` value must be lower or equal to `high` value
```
but the roll still plays after clicking through the error.
- Similarly, choosing in-app expression for our lone 65-note roll also prompts a non-breaking popup error.

broadwell:
> https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments

I like the color-coded highlighting!

Another bug related to the in-app expression settings: the values displayed for Welte Licensee rolls are actually the defaults for Welte Red rolls, although poking at the variables suggests that the actual values being used are the correct ones (i.e., the Welte Licensee default expression settings). Probably the interaction of the expression settings stores and the Expression Settings panel UI is being complicated by the Licensee expression box's inheritance from Red Welte?

Also it looks like `extendControlHoles()` is only in effect for 88-note rolls at the moment; it wasn't universally applied in earlier branches either (where it was usually called `applyTrackerExtension()` -- the new name is better until/unless we start using it to extend note holes), but really there's no reason for it not to be be used for every roll type with in-app expression. There's also considerable potential for it to be DRYed up so that it doesn't need to be implemented in full in every expression box.

Relatedly, I agree that `computeDerivedExpressionParams()` potentially could be DRYed up between the Duo-Art and 88-note expression boxe.

broadwell:
[8e9d8fe](https://github.com/sul-cidr/pianolatron/pull/126/commits/8e9d8fed040fe1b101d66a2188f913bb6e5948b7) is definitely a helpful refactoring, although it seems to have broken the hole overlays for all roll types other than Red Welte (maybe related to the fact that it's the only roll type that has downward panning).

broadwell:
Following up on my previous note: the final commit (https://github.com/sul-cidr/pianolatron/commit/c609233151e44435c44faefc03ba6c48cd15a383) in the temporary [`expressionizer-classes-hole-overlays-fix`](https://github.com/sul-cidr/pianolatron/tree/expressionizer-classes-hole-overlays-fix) branch restores the hole overlays/highlighting for non-Welte red rolls that is broken in https://github.com/sul-cidr/pianolatron/pull/126/commits/8e9d8fed040fe1b101d66a2188f913bb6e5948b7.

There's probably a way to implement this fix that aligns more elegantly with the refactored use of the `$holesIntervalTree` store rather than passing around the `holesData` structure, but I haven't been able to find it. So I figured I should at least share this (not so) quick-and-dirty correction in the hopes that it might be a step on the way to a more polished fix.

broadwell:
As noted above, the [`expressionizer-classes-fixes`](https://github.com/sul-cidr/pianolatron/tree/expressionizer-classes-fixes) branch includes proposed fixes for a handful of remaining bugs in this draft PR. It could be opened as a PR against this branch, but I've held off on doing that until it has been looked over by someone other than me. Here's a quick rundown of the issues addressed by its constitutive commits:

https://github.com/sul-cidr/pianolatron/commit/c609233151e44435c44faefc03ba6c48cd15a383: currently, the hole overlays only work for Welte red rolls. This extends the functionality to all roll types, although at the cost of much of the elegance introduced in https://github.com/sul-cidr/pianolatron/commit/8e9d8fed040fe1b101d66a2188f913bb6e5948b7 (there might still be a better way to do this).

https://github.com/sul-cidr/pianolatron/commit/f60fd7cebef7a6aadcd7100cd0ce4136b1c953c1: this restores the functionality of the "Reset to Defaults" button in the Expression Settings menu when in-app expression is enabled.

https://github.com/sul-cidr/pianolatron/commit/986056ce3511c80ff9f955dff04f1b936cf66ecf: a small tweak so that the "Hole visualization data is not available for this roll" message only appears when it's appropriate, instead of all the time.

https://github.com/sul-cidr/pianolatron/commit/13ca0a03405df915400682dc8e18d9216b1a5813: previously, the roll-specific expression settings were not being fully updated when a roll of a different type was selected with in-app expression already enabled, resulting in the new expression box being in an inconsistent state. This fixes that.

https://github.com/sul-cidr/pianolatron/commit/95226a50dedc98206548698f973ee2d91ca4ecc4: this builds on the previous commit and removes a now-unnecessary reactive variable subscription that also was previously causing some pretty serious performance slowdowns when changing in-app expression parameters.

PR #129, opened 2023-09-22
Upgrade Openseadragon dependency

broadwell:
Only one small code change is needed; tiled source loading performance is allegedly improved.


simonwiles (approved):
LGTM 👍️

PR #130, opened 2023-09-26
Dependency Updates

simonwiles:
This PR includes a bunch of dependency updates, most notably the upgrade from Astro `1.x` to `3.x`.  This necessitates various other bits and pieces, including a switch to Vite `4.x` and defaulting to ESM instead of CJS for module resolution.


broadwell (approved):
This all works for me!

PR #137, opened 2023-10-04
Add Listener / Performer UIs

cfitz:
Adds some very basic listener/performer profile functionality

Listener: 
- /?druid=qm069fy5200 
- Locked into a single roll ( no roll search/changing pulldown )
- Few controls ( volume, temp, start, stop ) 
- No welcome screen
- No measuring ruler

Performer:
- /performer/?druid=qm069fy5200
- Just the previous iteration of the UI, but in a new namespace

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -87,10 +87,10 @@
 
   import catalog from "./config/catalog.json";
 
-  export let profile = "performer";
+  export let profile = "perform";
 
   const joinPath = getPathJoiner(import.meta.env.BASE_URL);
-  const isPerformer = getProfile(profile) === "performer";
+  const isPerform = getProfile(profile) === "perform";
 
   let firstLoad = true;
   let appReady = false;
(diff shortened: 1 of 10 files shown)


broadwell:
This is great! But could this PR be made against `develop` instead? `main` is essentially a dead branch until we launch the app, so the idea is that `develop` is where the newly minted features should be merged first; then eventually they'll be added into `staging` as the more stable preview version, which eventually will be mirrored into `main` when we launch. These tiers are a bit cumbersome, but they may prove worthwhile, and anyway we do now get a nice auto-generated preview link from Netlify whenever a PR is made against `develop`.

Also, I think it's preferable to use `/perform/` in the URL, rather than `/performer/` (as I believe you also mentioned).

broadwell:
Thanks, I think this is an excellent "first cut" at dividing the UI entry points.  
Regarding the default listen view, eventually we'll need to make sure the tempo ratio readout is reunited with its slider rather than stuck at the bottom of the panel, but the look and feel of that panel is likely to change quite a bit in the future anyway (for one thing, we might also remove the key mappings above the slider), so we can leave it as is for now.

Issue #132, opened 2023-09-28
Create a front/landing page for the app, with text and links derived from the wireframe


broadwell:
As the wireframe suggests, the page can be a one-off in terms of its layout/design elements, though it should share some thematic similarities with the rest of the app pages and the player. It's fine to use the text and link ideas from the wireframe, but a few things to keep in mind:

- listen, perform, and explore can be featured links, but there probably won't be a high-level link for "analyze"
- we may not actually use "Pianolatron" for the title of this page. "Pianolatron" may eventually only refer to the performance view of the app; the player view would just be the "Piano Roll Player". But it's fine to use "Pianolatron" as a placeholder for whatever we decide to call the app as a w hole.

PR #141, opened 2023-10-05
Creates homepage (to develop this time)

markpernotto:
Not sure why I couldn't update my previous PR to just point to `develop`, but here we are. :)

Addresses issues raised in [this](https://github.com/sul-cidr/pianolatron/issues/132) ticket.

A few callouts:
- While we're targeting [this](https://www.figma.com/file/UL1tt48wgqCoXd6BCucq6t/pianolatron?node-id=1%3A5&mode=dev) Figma design, this can be fluid. This is only seen as a version 1 of the homepage. Links don't go anywhere, the URL isn't correct. 
- One new template and one new component were created. A new page was created, but ideally this would replace whatever the domain is pointing to currently

Screenshot 2023-10-02 at 9 51 23 PM

To access this page in the app, visit: `localhost:8080/homepage/`


broadwell:
This looks great! While we're at it, though, we might as well apply some changes that came up in conversations with the expert users but didn't find their way into the wireframe, and a few other tweaks that anticipate upcoming features:
- Let's change the text under "Pianolatron" in the upper right to read "Piano Rolls in the Stanford University Archive"
- The "listen" button should have a "to" in front of "piano rolls"
- The "listen" button can link to `/?druid=qm069fy5200` for now
- The "explore" button can link to `/search/`
- The "perform" button can link to `/perform/?druid=qm069fy5200` for now
- The "analyze" button should be replaced with one in the same style that reads "LEARN <newline> about piano rolls" and links to `https://exhibits.stanford.edu/supra`
- Let's change the blurb to read "and you can control it in real time with a game controller or computer keyboard to give your own expressive performance."

PR #143, opened 2023-10-11
Adds Static Header/Footer

markpernotto:
This adds a `stanfordHeader`, `stanfordSubHeader` and `stanfordFooter` to layouts, adds three fonts and (very loosely) follows the Stanford style guide found [here](https://identity.stanford.edu/digital/web-design/best-practices/).

This PR also allows for Markdown inclusion with the addition of a Markdown file, specifying which template that should be used: 
1. Template `markdownLayoutFullHeader` includes the header, subheader and footer
2. Template `markdownLayoutSingleHeader` includes the header and footer only (no subheader)

Player header:
Screenshot 2023-10-27 at 9 44 41 AM

Full header:
Screenshot 2023-10-27 at 9 45 10 AM

Single header:
Screenshot 2023-10-27 at 9 50 43 AM

Footer:
Screenshot 2023-10-27 at 9 45 27 AM


src/layouts/base.astro
```diff
@@ -30,6 +30,6 @@ const datetime = new Date().toISOString();
   </head>
   <body>
     <slot />
-    <Footer />
+
```
markpernotto, line 57: Unsure if this is a good idea, but needed to see the `stanfordFooter` properly.

src/pages/generic.astro
markpernotto, line 1: This page was created for the sole purpose of using it to test out the header and footers created - it should probably not be committed.

src/styles/base.scss
```diff
@@ -56,62 +56,62 @@ h1 {
...
+
+//     path {
+//       fill: white;
+//     }
+
+//     rect {
+//       fill: grey;
+//     }
+//   }
+// }
```
markpernotto: Leaving this commented out to highlight the issue(s) of:
1. Do we want global footer styling here? Maybe not since nothing outside of the footer elements would be affected?
2. Is there anything in here we _should_ retain for our existing footer?

src/styles/fonts.scss
```diff
@@ -8,3 +8,19 @@
...
+@font-face {
+  font-family: SourceSans3;
+  src: url("fonts/SourceSans3-VariableFont_wght.ttf") format("truetype");
+}
+
+@font-face {
+  font-family: Stanford;
+  src: url(https://www-media.stanford.edu/assets/fonts/stanford.woff) format("woff"),url(https://www-media.stanford.edu/assets/fonts/stanford.ttf) format("truetype");
+  font-weight: 300
+}
```
markpernotto: This is not explicitly defined in the design spec; rather, it's used to achieve the home logo. If everyone likes this, we can also have access to this font as well, even though it's hosted. (I do believe we can download this font as well - I will check on that today)

src/styles/base.scss
broadwell: I think the player/listener app should have its own minimalist footer that will be pretty similar to the current one for the app (single row, mostly transparent except for the clickable parts), while the static pages will have the big Stanford-style footer. So probably the styles for these should live in different places?

broadwell: There's also the option of layering the Pianolatron-specific footer above the Stanford footer on the static pages, similar to how the Stanford Libraries footer is layered above the Stanford footer, but I'm not sure that will be necessary, especially as all of the links from the Pianolatron footer will probably be linked from other places (like the header) on the static page template.

src/layouts/stanfordFooter.astro
```diff
@@ -0,0 +1,191 @@
...
+          .su-global-footer__menu {
+            font-size: 1.25rem;
+            list-style-type: none;
+            margin: 0 0 1rem;
+            padding: 0;
+
+          }
+          .su-global-footer__menu li {
+            display: inline-block;
+            font-family: SourceSanse3, sans-serif;
```
broadwell: Probably this should be `SourceSans3` ?

markpernotto: Fixed! Thank you!

src/styles/base.scss
markpernotto: So, I've kind of introduced a concept of a Header and then a SubHeader. We only have one Footer now, but that's not to say we couldn't add a SubFooter.

I think, for the purposes of this code, though, we've come to the conclusion that we don't need footer styling here and this is okay to remove?

broadwell:
Just a few follow-ons as discussed recently:

- Firefox somehow still doesn't support nested CSS, so the header and footer don't show up properly there (though they're fine in Chrome/Edge). When I had to deal with this for a different task, I used [postcss-cli](https://www.npmjs.com/package/postcss-cli) to unroll nested CSS.
- While we're at it, I think it looks better for the bottom copyright line of the Stanford footer to be white, like the rest of the text.
- The preferred links for the subheader:
  - "HOME" links to /homepage/ (we may change the URL of this page later)
  - "ROLL ARCHIVE FEATURES" can be changed to "ROLL ARCHIVE" and links to https://exhibits.stanford.edu/supra
  - "BROWSE" should be changed to "SEARCH" and points to the app's search page, ideally at /search/
  - "ABOUT" can point to this page (/about/)
- It would be great to have a template for the player that just shows the app-specific subheader, with no Stanford header or footer, although this could be a separate PR.

markpernotto:
> Just a few follow-ons as discussed recently:
> 
> * Firefox somehow still doesn't support nested CSS, so the header and footer don't show up properly there (though they're fine in Chrome/Edge). When I had to deal with this for a different task, I used [postcss-cli](https://www.npmjs.com/package/postcss-cli) to unroll nested CSS.
> * While we're at it, I think it looks better for the bottom copyright line of the Stanford footer to be white, like the rest of the text.
> * The preferred links for the subheader:
>   
>   * "HOME" links to /homepage/ (we may change the URL of this page later)
>   * "ROLL ARCHIVE FEATURES" can be changed to "ROLL ARCHIVE" and links to https://exhibits.stanford.edu/supra
>   * "BROWSE" should be changed to "SEARCH" and points to the app's search page, ideally at /search/
>   * "ABOUT" can point to this page (/about/)
> * It would be great to have a template for the player that just shows the app-specific subheader, with no Stanford header or footer, although this could be a separate PR.

I believe I've fixed all of the above, but please double-check me?

broadwell:
This has come together really well -- thanks for slogging through all the spec refinements.
I still have one pretty big change request re: the player page, but I'll make a separate ticket for it after merging this.

PR #148, opened 2023-10-19
ItemsJS Library

markpernotto:


broadwell:
Cool, this looks good, too! Hmm...
The facets/filters on the left side with the tallies of the matching elements is a nice feature -- not strictly necessary, but nice. Any idea whether the search piece can be configured to do expansion from ASCII characters to their diacritical forms out of the box? That's close to an essential feature given the large numbers of umlauts and accents in the roll data.
Besides coverage of the essential features, our final choice of a framework for the search and browse page also may come down to factors like perceived maintainability, ease of integration into the rest of the codebase, and how much it adds to the JS download bundle.

broadwell:
Moving this to draft status until we get the chance to come back to it and determine whether there are ways to get the features we want out of it while also retaining the best features that the GridJS implementation provides.

broadwell:
Moving this to archival status.

PR #149, opened 2023-10-19
some quick svelte stuff

cfitz:


src/components/Search.svelte
```diff
@@ -109,48 +109,35 @@
     ],
   };
 
-  export const compositions = itemsjs(catalog, configuration);
```
cfitz, line 112: You only need to use "export" for properties you need to pass in, so there's really no point using that for something like this right now..

src/pages/search.astro
```diff
@@ -4,5 +4,5 @@ import Search from "../components/Search.svelte";
 ---
 
 <BaseLayout title="Pianolatron">
-  <Search />
```
cfitz, line 7: This was probably your biggest problem right here...it's a funny one..
We are only using the client-side version of Svelte, so you need to indicate that here...you on:clicks were probably firing on the server-side, which itemjs probably has no idea about...

src/components/Search.svelte
```diff
@@ -109,48 +109,35 @@
...
-      return results;
-    }
+  const textSearchCompositions = ( {target } ) => searchCompositons("composer_desc", target.value);
+	onMount(() => {
+    searchCompositons("composer_desc");
+    console.log(results);
   })
 
-
+  $: results;
```
cfitz, line 129: ignore this! Pretty sure this isn't doing anything...

src/components/Search.svelte
```diff
@@ -109,48 +109,35 @@
...
+  
 </script>
 
 <h1>Super Simple Sorted List</h1>
-  <input on:change={(e) => {
-    textSearchCompositons(e.target.val)
-  }} placeholder="Search" />
+  <input on:input={textSearchCompositions}
```
cfitz, line 134: you could also just do this on:input={ (event) => testSearchCompositions(event) }

markpernotto (approved):

PR #161, opened 2023-11-03
Adds Styling to Player Buttons

markpernotto:
This adds some simple styling to the player icons for the app. Probably still a work in progress, would appreciate feedback!

untouched
Screenshot 2023-11-02 at 5 00 17 PM
hover over Play button
Screenshot 2023-11-02 at 5 02 10 PM
hover over Repeat button
Screenshot 2023-11-02 at 5 01 55 PM
Record turns red when pressed
Screenshot 2023-11-02 at 5 00 34 PM
play (pause) button stays white while playing
Screenshot 2023-11-02 at 5 00 25 PM

### DIFF
--- a/src/components/RollPlayerControls.svelte
+++ b/src/components/RollPlayerControls.svelte
@@ -1,11 +1,8 @@
 <style>
-.player-buttons {
+.player-button-container {
     padding: 2px;
     text-align: center;
 }
-
-
-
 </style>
 
 <script>
@@ -54,9 +51,9 @@
 
 </script>
 
-<div class="player-buttons">
+<div class="player-button-container">
 <IconButton
-    class={ $playRepeat ? "overlay" : ""}
+    class={ $playRepeat ? "overlay player-button" : "player-button"}
     disabled={false}
     on:mousedown={togglePlayRepeat}
     iconName="replay"
@@ -65,7 +62,7 @@
     width="24"
   />
 <IconButton
-    class=""
+    class={"player-button"}
     disabled={false}
     on:mousedown={resetPlayback}
     iconName="rewind"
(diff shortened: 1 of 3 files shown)


cfitz:
Looks good!

broadwell:
This looks great to me. @cfitz feel free to merge it into #151 if you like. A couple of further requests:

- We'll also need to add on/off indicator stylings to the repeat button eventually -- probably it should just be white while selected.
- And I think the start/end boundary selection buttons should have a set/clear paradigm: the first click sets the boundary and turns the icon white, the second click clears the boundary and turns the icon back to gray.
There's probably no need to indicate visually when one of those buttons is not usable -- if for example the user tries to set a start boundary that's beyond the end boundary, it should just fail silently (same when the user tries to set an end boundary that's before the start boundary).
- Let me know if you have thoughts about the above or if I'm missing a corner case or something more fundamental (which is quite possible).

markpernotto:
> This looks great to me. @cfitz feel free to merge it into #151 if you like. A couple of further requests:
> 
> * We'll also need to add on/off indicator stylings to the repeat button eventually -- probably it should just be white while selected.
> * And I think the start/end boundary selection buttons should have a set/clear paradigm: the first click sets the boundary and turns the icon white, the second click clears the boundary and turns the icon back to gray.
>   There's probably no need to indicate visually when one of those buttons is not usable -- if for example the user tries to set a start boundary that's beyond the end boundary, it should just fail silently (same when the user tries to set an end boundary that's before the start boundary).
> * Let me know if you have thoughts about the above or if I'm missing a corner case or something more fundamental (which is quite possible).

Hi @broadwell thanks for pointing those out. I think each of those could be put into bug tickets we can easily come back to later.

broadwell:
> Hi @broadwell thanks for pointing those out. I think each of those could be put into bug tickets we can easily come back to later.

Yeah that seems like a good way to handle the follow-on requests. For now, either of you can free to go ahead and merge this PR unless you would prefer me to do it.

cfitz (approved):

PR #162, opened 2023-11-05
Search/Browse Page using Gridjs

cfitz:


broadwell:
This is great! Could we make a couple of further enhancements before merging?

1. Change the heading of the "Label" column to "Title"
2. Replace the "Type" column with one for the "Publisher" data

It'll be good to merge once the above are addressed, although these follow-up items also will need to be addressed soon:
1. Change the name of the page and the link in the site header to `/search/` and "Search" (rather than Browse)
2. Integrate this search component into the site template with the "Home | Roll Archive Features | ..." header at the top and the Stanford footer at the bottom.

cfitz:
Made the changes, except for the header/footer. Sorry is that in place? I couldn't find it

Do we want to add the HTML "hit highlighting" in the results?

broadwell:
Great, thanks! And yes, I'm in favor of turning on the hit highlighting (at least on a trial basis) before merging.

I think this would be the first case of a non-markdown page to have all of the header and footer elements, but it _should_ just be a matter of importing the `<StanfordHeader />`, `<StanfordSubheader />` and `<StanfordFooter />` layouts and adding the elements in the appropriate places within the `<BaseLayout />` in `search/index.astro`

cfitz:
Ah, realized i needed to pull down deploy to get the headers/footers. 
Added those now and changed the page to "Search"

Added hit hightlights... but need to get the HTML to render in the grid ( right now you just see <mark> tags in there )
Also noticed i need to add something to remove the tags if user deletes the search query from the input.

Dunno, I am still not really excited about the way the pages looks though?

broadwell:
I'm still seeing a lot of weirdness that I think is due to the hit highlighting not quite being dialed in yet -- one example is screenshotted below. If the hit highlighting is proving to be a lot trickier than expected, then I think it's fine to merge this PR without it and maybe add it later.
We also can continue to iterate on the look and feel of the page, but we may be getting some feedback on this and other new/drafted features soon from other stakeholders, so I think it's fine just to merge it as soon as the hit highlighting issues are resolved or we decide to drop that for now.
Screen Shot 2023-11-09 at 9 30 21 PM

cfitz:
Yeah, sorry a late night push that I missed a couple of things...I realized it was using "Title" field for the search instead of the "Work" field, so all the highlights were offset some and not correctly marking the right term.  
I fixed that and added the "Arranger" and added the "Number" field to the end of publisher.

PR #163, opened 2023-11-08
Cleans up Left Panel

### LINKED ISSUE #154: Clean up the metadata display in the left pane
There are only a couple of high-priority changes here:
- Remove the "Call Number" field entirely
- Hide the "Performer" field entirely when this value is missing from the metadata (e.g., 88-note rolls never have performers, because they were programmed in the factory, not recorded from an actual person playing the piano)
- (Stretch goal): The composer, performer, and publisher names all should link to the /search/ page (opening in a new tab when clicked) with the name autofilled in the search bar and the search results already filtered.

This task also could include other improvements to how the left-side pane looks. One basic refinement I'd suggest:
- Making sure the "Copy bookmark" button looks nice and matches stylistically with the rest of the UX, and functions properly of course.
- Feel free to suggest others!

markpernotto:
This cleans up some of the data in the left panel and adds some additional data as well.

This addresses issues raised in this ticket: https://github.com/sul-cidr/pianolatron/issues/154

### DIFF
--- a/src/components/RollDetails.svelte
+++ b/src/components/RollDetails.svelte
@@ -44,10 +44,35 @@
   dd :global(span) {
     opacity: 0.5;
   }
+
+  ul {
+    margin: 0;
+    padding: 0;
+    list-style: none;
+  }
+
+  li {
+    line-height: 1rem;
+    padding: 0 0 10px;
+  }
+
+  li a {
+    font-size: 1rem;
+    text-decoration: none;
+  }
+
+  li a:hover {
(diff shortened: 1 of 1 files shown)


broadwell:
Thanks, this looks great! Definitely an improvement on what was there before.
There probably will be further refinements that come to mind down the road, and also there's more work to be done to integrate the roll bookmarking feature, assuming the button for that will live on the left panel, but let's get this in the code base as a new foundation for the panel design.

PR #166, opened 2023-11-13
155 integrate the playback lag warning

### LINKED ISSUE #155: Integrate the playback lag warning
We can tell when the user's browser isn't keeping up with the tasks of displaying the roll and playing the notes on time by calculating in advance the millisecond when each note *should* be played, and then comparing this to the time (counted in ms from when the user clicks "play") when each note actually gets played. When this value exceeds a certain threshold, we should display a roll-down warning (like the one used to ask the user "are you sure" when a large number of samples is chosen in the piano menu) that says something like:

"Your web browser seems to be having difficulty playing this roll, and as a result some notes may not be sounding when they should. Please consider following the suggestions **in the documentation** to attempt to improve the performance."

(**in the documentation** will be a target="blank" link to a static documentation page that we'll write later).
 
There is already code in the [playback-slowdown-warning](https://github.com/sul-cidr/pianolatron/tree/playback-slowdown-warning) branch to detect when the playback is getting too slow, but it may need to be cleaned up and updated, and this detector also needs to be linked to the roll-down warning popup as described above.

cfitz:
Starting point for the latency warning. 
The SamplePlayer keeps a list of notes that lagged. If there's a certain number ( 10? ) within a relatively recent period ( 1% of the total ticks ) of the current tick, we consider this enough latency to do something ( right now, put a message in the console ).

Just guessing about the numbers at this point. 

Using Chrome Dev Tools -> CPU Throttle -> 6x slow down should trigger the warning ( cranking up tempo and disabling Chrome's access to GPU also helps ).

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -97,10 +97,12 @@
     playExpressionsOnOff,
     rollPedalingOnOff,
     playbackProgress,
+    showLatencyWarning,
   } from "../stores";
   import { clamp, getHoleLabel } from "../lib/utils";
   import RollViewerControls from "./RollViewerControls.svelte";
   import RollViewerScaleBar from "./RollViewerScaleBar.svelte";
+  import LatencyWarning from "../ui-components/LatencyWarning.svelte";
 
   export let showScaleBar = true;
   export let imageUrl;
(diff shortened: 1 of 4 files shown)


cfitz:
Added some UI, which was also just guessing of how it should look. Took the "Loading Roll" UI and added some text to it. 
Also not sure if the warning triggers too soon or if it takes too long to turn off? The way it's implemented, it depends on the tempo that the roll is playing, so for some of the slower roles, the warning might drag on too long?

PR #178, opened 2023-11-27
Move "Copy URL Button" to control bar. Make start/stop buttons toggle

### LINKED ISSUE #173: "Mark Start" and "Mark End" buttons should be toggle-able
Ideally, each button icon would be styled active (white) when it has been used to set a start or stop point. Clicking the button again would remove that start or stop point (and remove the line from the viz) and change the button style back to the default gray.

A couple of related requests:
- The tooltip of the "Mark End" button currently reads "Mark Start" - would be great to fix this.
- Can we allow the user to mark 0% as the start point? This is entirely redundant because 0% is the start point when no start point is selected, but still, the user might be confused about why they can't set a start point at 0%.

### LINKED ISSUE #175: Move, restyle and possibly merge "Copy URL" and "Copy URL With Timestamp" buttons
My preference for this (at the moment) is actually to replace both buttons with a single bookmark button icon at the far left of the control strip, to the left of the "repeat" button.
In this scenario, the position of the play point won't make any difference to the URL. Instead, the only way to set the start timestamp in the URL is to use the "Mark Start" button prior to clicking the bookmark button (otherwise, the start point is assumed to be 0%). Similarly, if the "Mark End" button has been used, this value also will be included in the URL via the appropriate query parameter.

I'm certainly open, however, to ideas that involve having more than one bookmark button. But in any case, the current positioning and appearance of the buttons isn't ideal, so it would be great to improve them.

cfitz:
This moves the Copy URL button to the control bar, with a "bookmark" icon. The current progress time of the roll is NOT used in the start param, but only the playbackProgressStart.

playbackProgressStart and End now toggle the markers.
It should be possble to set the marker at "0"


broadwell:
This is great -- the little acknowledgment animation is especially helpful. There are aspects of the functionality and certainly the UX that could be fine-tuned, but I think it's best to merge this into `develop` right away.

PR #182, opened 2023-11-28
169 remove currently playing roll from similar works by this performer list

### LINKED ISSUE #169: Remove currently playing roll from "Similar works by this performer" list

cfitz:
Add the druid to the roll Metadata, which will be filtered out from the "Similar Items By Composer" list


broadwell:
This works well for me. We may eventually want to change the links to related rolls while in the Perform mode to also point to the pages for those rolls in Perform mode (currently they always point to the Listen interface), but that's for another time.

PR #185, opened 2023-12-04
Move Player Buttons for Perform View

### LINKED ISSUE #156: Move and restyle soft, sustain, accent buttons/indicators
The soft, sustain, and accent buttons are removed entirely from the listen view (the animated pedals will still indicate when soft and sustain are being used), and the accent button has no use in this view.

In the perform view, these buttons should be moved elsewhere and restyled. Feel free to use your discretion when choosing the best place in the layout to position them.
Consider also the [better-playback-controls](https://github.com/sul-cidr/pianolatron/tree/better-playback-controls) and [indicators](https://github.com/sul-cidr/pianolatron/tree/indicators) branches for ideas.

markpernotto:
This PR re-introduces the player buttons for the perform view, but with some more UX polish.
Addresses issues presented in this ticket: https://github.com/sul-cidr/pianolatron/issues/156

Screenshot 2023-12-03 at 10 12 56 PM
Screenshot 2023-12-03 at 10 12 34 PM
Screenshot 2023-12-03 at 10 12 23 PM
Screenshot 2023-12-03 at 10 12 13 PM

This has been done with baseline support from @simonwiles `better-playback-controls` PR here: https://github.com/sul-cidr/pianolatron/tree/better-playback-controls
as well as some feedback from other stakeholders during this process.

*Note - I'm not entirely sure when `isPerform` should be true - I also made a couple of other comments regarding features not yet build out.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -45,7 +45,7 @@
     bottom: 0;
     width: calc(100% - var(--navigator-width));
     opacity: 0.5;
-    z-index: z($main-context, keyboard-overlay);
+    // z-index: z($main-context, keyboard-overlay);
   }
 </style>
 
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -137,10 +137,8 @@
 
   // redundant, but the way the BasicSettings comp is built requires we define the func
   // here, as it won't update the ref.
-  const skipToPercentage = (percentage = 0) => {
-    console.log("skipping");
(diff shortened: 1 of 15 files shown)


broadwell:
As noted in the final commit, it's looking unlikely that we'll use most of the restyled buttons in the right-hand panel, leaving just the pedal and accent indicators/buttons there, but hopefully the changes made to the `<Icon>` and `<IconButton>` components can be useful when overhauling the control strip buttons above the roll viewer.

PR #189, opened 2023-12-10
minor nav sections cleanup

cfitz:
DRYs out the selection svg overlay code.


broadwell:
This looks good to me. There's an issue with the bookmarking functionality when a section is highlighted, but that's not relevant to this PR, so I'll add it to a new ticket.

PR #190, opened 2023-12-11
Updating and integrating the legend feature

### LINKED ISSUE #157: Consider integrating Simon's "legend" panel
The idea is to try to integrate as many of the new features from the PR https://github.com/sul-cidr/pianolatron/pull/86 as we can.

At the least, we should display the instantaneous roll velocity (in ft/min^2) in the perform view. Showing a legend for the hole colors and color ranges in the perform view is also a nice-to-have.

The legend code is already pretty functional, but may need a final once-over, and as it's fairly old, some parts may need to be updated to match up with how the app currently works.

cfitz:
Updates  #86. 
Probably need to sort out some of the functionality, mainly where do we want this to be in the UI?


broadwell:
Thanks, this looks good! It might be preferable for the timecode to include sub-second values, but that can be finessed later.

PR #194, opened 2023-12-13
Bug with bookmark copy taking playbackProgress and not playbackProgre…

cfitz:
It was using the wrong variable ( playbackProgress and not playbackProgressStart )


broadwell (approved):
Looks good to me.

PR #195, opened 2023-12-13
188 fix startstop markers should reset when choosing a new roll from the drop down perform mode only

### LINKED ISSUE #188: Fix: start/stop markers should reset when choosing a new roll from the drop-down (Perform mode only)
Assuming we're going to keep the drop-down menu in the Perform mode (which I think is likely, though I'll ask the PI about the next time we meet), this causes some WTF but survivable behavior when changing to a roll of the same type, but is an app-breaking bug when switching from e.g. a Welte roll to an 88-note roll. It fails with the following error:
`Uncaught Error: `low` value must be lower or equal to `high` value`

cfitz:
This fixes the issue with start/stop markers persisting after using the roll selector drop-down.
Also adds some fixes for backwards rolls.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -221,7 +221,10 @@
 
     Promise.all([mididataReady, metadataReady, pianoReady]).then(
       ([, metadataJson]) => {
-        metadata = (({ holeData: _, ...obj }) => obj)({...metadataJson, druid: rollDruid});
+        metadata = (({ holeData: _, ...obj }) => obj)({
+          ...metadataJson,
+          druid: rollDruid,
+        });
         holeData = metadataJson.holeData;
         annotateHoleData(holeData, $rollMetadata, $scrollDownwards);
         buildHolesIntervalTree();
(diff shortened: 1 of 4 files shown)


broadwell:
Thanks, this does indeed reset the start/stop makers when switching to a new roll, avoids the breaking errors with "backwards" (unrolling top to bottom) rolls, and generally streamlines the behavior of this feature (which turns out to be quite tricky to work out completely).
I'm still noticing some issues with the backwards-playing rolls, though, particularly that the markers and region highlight in the navigation bar still seem to be displaced from their correct positions, and don't always clear properly when the markers are deselected a new roll roll is chosen (see screenshot below).
If it would be helpful to merge this change as is and then open a new ticket and branch to handle these remaining edge cases, we can do that, or the final fixes can go in as additional commits here -- just let us know what you prefer.
Screen Shot 2023-12-18 at 3 14 00 PM

cfitz:
Added a few more fixes for the backwards rolls. 
There's some weirdness with OSD on these, but i think I sorted out most of the issues.
The fix is a bit hacky, but includes passing an empty onDraw function to the addOverlay arguments, which prevents OSD from trying to calculate the overlay position ( which can be off since not all the DOM elements have been loaded ). This seems to improve the accuracy/consistency of the overlay location a lot. 

https://deploy-preview-195--pianolatron-develop.netlify.app/perform/?druid=xy736dn5214&start=0.40&end=23.74

Changing rolls sometimes resulted in random markers being added in the navigator, but I think I fixed those. Will keep playing with it to see if any other cases pop up

Issue #192, opened 2023-12-12
Search page refinements


broadwell:
- Add a persistent, highlighted (bold) "**Search:**" to the left of the search bar; remove the "SEARCH" placeholder text from inside the search bar.
- While we're at it, also add a "Roll type:" label to the left of the row of roll type filter selection buttons/pills
- Ignore `RETURN`/`ENTER` keypresses in the search field -- currently, typing `RETURN` in the search bar shifts the visible search text up and above the actual bar (because it moves the cursor down to the next "line"). Because the search results are updated whenever any other characters are entered in the field, I think we can safely just ignore `RETURN` and `ENTER` when they're typed in this field.
- Increase the number of results rows in a page of search results to 20 (rather than 10).
- Display appropriate icons rather than text for the `Play` and `Perform` links for each search result.
- The current/previous search text should persist in the search bar until explicitly deleted (currently it is cleared automatically every time the focus moves away from and back to the search bar.
- Remove the red rectangular (`active`?) highlight that appears around each button while and after it is pressed (this also happens for the header links, and also should go away there)
![Image](https://github.com/sul-cidr/pianolatron/assets/7329112/221f8feb-1d69-4791-a8b1-a1d911e03b91)

Stretch:
- Consider concatenating the "Composer" and "Arranger" fields so that the mostly empty "Arranger" column doesn't take up so much horizontal space in the results (this probably would need to be done when creating the `catalog.json` file).

PR #196, opened 2023-12-18
Search page bug fixes

cfitz:
Issues highlighted in ticket #192


broadwell:
These changes all look great - thanks! Putting the arranger on a separate line but in the same cell as the composer strikes me as a straightforward and elegant solution.

PR #198, opened 2024-01-22
Update Stacked Headers

markpernotto:
You can access it if you specify the `newheader/` URL

This can be used either as just a top header or use both to make a full header.

Not 100% sure all of the links are where we want them to be.

Will be addressing fonts in a different ticket.


cfitz:
Looks good to me!

PR #199, opened 2024-01-30
160 integrate recording functionality

### LINKED ISSUE #160: Integrate recording functionality
Ideally, both MIDI recording and audio (.wav) recording would happen at the same time when the record button on the transport control strip is activated (as long as this doesn't overload the browser). This would simplify the UX considerably, as the distinction between the two types of recordings would only be relevant at the final download step (and maybe we could just give the user both files then without asking which one they want).

There are branches with working code for both; [audio-recording](https://github.com/sul-cidr/pianolatron/tree/audio-recording) uses the standard browser WebAudio API and is pretty straightforward, but [midi-recording](https://github.com/sul-cidr/pianolatron/tree/midi-recording) relies on a [modified version of MidiWriterJS](https://github.com/broadwell/MidiWriterJS) (that is also not up-to-date with the upstream anymore) and would need some work to integrate into the app and build system.

A decision is yet to be made re: whether recording functionality should be available in the Listen mode.

cfitz:
#160

Adds commits for audio recording and enables record button in the control strip. 
When a recording is "paused", a notification is presented allowing the user to export, clear, or continue their recording. 
Export WAV file is druid + UTC timestamp ( 🤷 ) 
Keeps the original recording buttons in the midi control panel.


broadwell:
This is looking good to me. I really like the idea of putting the "Export/Clear" functionality in a popup dialog.

My only suggestion for now is that I think that if we're going to allow a recording to persist in the recording buffer until it's explicitly deleted (or maybe exporting could always trigger deletion, but that's another question), there should be some signal in the UI to show that the buffer is not empty even when not actively recording. Maybe the middle dot of the recording button could be red if there's something in the buffer, and the outer circle (which may just be the border of the entire button when we switch to the new icon set) would turn red while actively recording?

In any case, it **is** nice for the user to be able to pause and later resume recording without deleting the buffer -- it just introduces some complications to the UX that need to be handled.

cfitz:
Yeah, like the idea of having something to let user know they have a recording in buffer. Added the "keep the center dot red in record icon when there's a recording in buffer"

But also not totally satisfied with UX. I used the notification component because it was already there and it works well..but just need something more than assuming the user knows to go to the MIDI control pane to do some tasks, but unsure what would work best.

cfitz:
This should be superseded by #203

PR #200, opened 2024-01-30
Fix disapearing control strip in Chrome

### LINKED ISSUE #168: Fix: Manually click-dragging the roll down hides transport strip in some situations
"Dragging the roll down" actually involves clicking on the roll image and dragging the mouse pointer **up** (so that the viewport pans down the roll).
I've noticed this on Chrome (for Mac) when the piano keyboard is **visible but not maximized**. Could not reproduce with Firefox.

cfitz:
Fixes weird issue when dragging the roll the viewer jumps up and pushes the control strip behind the header.
Only in Chrome.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -26,7 +26,8 @@
 
   #roll {
     position: relative;
-    flex: 1 0 auto;
+    display: flex;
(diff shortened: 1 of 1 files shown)


broadwell:
Thanks -- this did the trick!

PR #203, opened 2024-02-15
Add WAV and MIDI Recording

cfitz:
Unifies WAV and MIDI recording in the performer view. 
Users can record, then when the pause are presented with a dialog to export WAV, MIDI, Clear their recording, or do nothing ("continue recording") 
Additional UI is still in the MIDI setting pane, although this can probably be removed at some point


broadwell:
Thanks, it's great to have this functionality in the `develop` branch. There are maybe still some UI nuances to work out, but I think it's definitely preferable to do both MIDI and WAV recording at the same time.

PR #212, opened 2024-05-14
More a11y hoop jumping

simonwiles:
Minor a11y tweaks.

### DIFF
--- a/src/components/Search.svelte
+++ b/src/components/Search.svelte
@@ -97,7 +97,7 @@
   // makes it very difficult to reuse that component here.
   const icons = {
     piano: `
-      <svg height="24" width="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round" name="piano" tabindex="0">
+      <svg height="24" width="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round" name="piano">
         <rect x="2" y="7" width="20" height="12" rx="2"></rect>
         <line x1="7" y1="9" x2="7" y2="13.5" style="stroke-width:2.5;stroke-linecap:square"></line>
         <line x1="12" y1="9" x2="12" y2="13.5" style="stroke-width:2.5;stroke-linecap:square"></line>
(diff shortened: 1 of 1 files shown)


src/components/Search.svelte
```diff
@@ -97,15 +97,15 @@
   // makes it very difficult to reuse that component here.
   const icons = {
     piano: `
-      <svg height="24" width="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round" name="piano" tabindex="0">
```
broadwell, line 100: I put the `tabindex="0"` there because I couldn't focus those links via keyboard tabbing in Firefox on Mac OS otherwise (though it worked fine in Chrome). Turns out this was due to FF following the Mac system configs, which by default disable link focusing via keyboard, for some reason:
https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox

Slightly related: the WAVE a11y checker still complains that those icon links have "NO ACCESSIBLE NAME" but it doesn't flag this as an error, so I guess we'll just go with it.
Screen Shot 2024-05-14 at 5 45 59 PM

src/components/Search.svelte
```diff
@@ -97,15 +97,15 @@
   // makes it very difficult to reuse that component here.
   const icons = {
     piano: `
-      <svg height="24" width="24" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round" name="piano">
```
broadwell: Yes this was left incomplete previously. Worse yet, the icons are actually incorrectly named (`piano` should be `play` and vice versa), which prevents one from choosing the right `aria-label` and `name` attributes. I'll just merge this and add another commit to fix the original error.

simonwiles: Oops, sorry, I thought I'd fixed that -- thanks for catching it!

src/components/Search.svelte
simonwiles: That's... odd -- WAVE on my machines here (as a Chromium extension) *doesn't* complain now, but did before...

`tabindex="0"` would be cool on the `<a/>` if necessary, I think?

broadwell: Yeah, I don't think it's necessary, though. It seems that the accepted solution in this case really is just for me to go into my system settings and enable link focusing via keyboard navigation.
Also those messages from WAVE aren't actual errors, and considering that the `<a/>` elements do have `title` attributes, I think they can be disregarded.

PR #213, opened 2024-05-15
Rewrite of the search table component

simonwiles:
This PR does away with the `gridjs` and `gridjs-svelte` dependencies and just builds a table with Svelte.  This was prompted partly by some a11y-checker false positives generated by the `gridjs` implementation, but mainly the jittering of the table when it was updated (and the lack of control over it) was bugging me.

There are also a couple of minor, mostly-unrelated, tweaks to the page layout template CSS.


broadwell:
This all looks great to me. Thanks!

PR #216, opened 2024-05-22
Accessibility accommodations for search page

broadwell:
These are all of the accessibility fixes I can come up with based on the feedback we've received so far. The commit messages tell the whole story, but here's a summary of the most significant changes:
- Changed the layout of the facet pills so they don't push the "Filtered: " text off the edge of the screen when zoomed at 200% on a laptop screen.
- Added logic to set the `aria-sorted=ascending|descending` attribute on table column headers
- Expanded the link button labels and titles so that instead of just reading "Play roll" they now read "Play roll [TITLE]", etc.
- Swapped the Publisher/Label and Play/Perform/etc. links columns so that the leftmost data column can serve as row headers (with the appropriate markup), improving legibility of the table contents via screen readers.
- Various color scheme tweaks so that selected pills and active headers don't trigger minimum text/background contrast rules.

### DIFF
--- a/src/components/Search.svelte
+++ b/src/components/Search.svelte
@@ -149,6 +149,7 @@
         &:hover,
         &:focus {
           background-color: #e5e7eb;
+          color: black;
         }
       }
     }
--- a/src/components/Search.svelte
+++ b/src/components/Search.svelte
@@ -55,7 +55,7 @@
       li {
         display: inline-block;
         border-radius: 6px;
-        background-color: darkslategray; // var(--primary-accent);
+        background-color: darkslategray;
         color: white;
         padding: 1px 8px 4px;
         margin: 0 4px;
@@ -77,9 +77,18 @@
     }
   }
 
+  .row-links-cell {
+    padding: 0;
+  }
+
+  .row-links {
+    display: flex;
+    align-items: center;
+    justify-content: space-evenly;
+  }
+
   .row-links a {
     color: var(--primary-accent);
-    display: inline-block;
     height: 20px;
 
     :global(svg) {
@@ -100,7 +109,7 @@
     text-align: left;
     width: 100%;
 
-    th {
+    th:not([scope="row"]) {
       background-color: #f9fafb;
       border: 1px solid #e5e7eb;
       box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


src/components/Search.svelte
```diff
@@ -46,7 +46,7 @@
       }
 
       ul {
-        flex: 1 0 auto;
+        flex: 1 0 min-content;
```
simonwiles, line 51: FWIW I can't get this to clip with or without this change, even up to 250% on FF or Chromium, but *with* this change it does start to wrap unnecessarily and "uncleanly" at about 230%.

250% before:
image

250% after:
image

I don't think we can worry about > 200%, and presumably this change *does* improve affairs in environments you tested, but I might just add a `gap-y` in an upcoming PR anyway, for if it does wrap at all.

src/components/Search.svelte
```diff
@@ -147,18 +158,20 @@
     td:first-child,
     th:last-child,
     td:last-child {
-      width: min(16vw, 240px);
+      width: min(17vw, 240px);
```
simonwiles, line 163: To be clear, this just shifts the viewport width at which "Image" gets ellipsized (is that a word?) from ~1400px to ~1320px (should always be using the webfont, so default fonts shouldn't matter).

src/components/Search.svelte
```diff
@@ -401,66 +414,67 @@
...
+          class:sortedDesc={sortOrder === "publisher-desc"}
+          >Publisher / Label</th
+        >
+        <th
+          tabindex="0"
+          scope="col"
           on:click={() => sortItems("work")}
           on:keypress={({ key }) => key === "Enter" && sortItems("work")}
+          aria-sort={(sortOrder === "work-asc" && "ascending") ||
+            (sortOrder === "work-desc" && "descending")}
```
simonwiles: The way I read the [`aria-sort` docs on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) I think the attribute should be absent when the column is unsorted, rather than `aria-sort="false"`:

> The aria-sort attribute should only be added to a single table or grid header at a time

I'm going to add a commit to this branch to implement this; shout if you think this needs more discussion.

simonwiles: I've been working through in commit order and just noticed you had this working (what I read as) "properly" and it got broken satisfying the linter...  I'll fix it in a way that keeps the linter happy.

simonwiles (approved):
One overall comment that doesn't belong in-line.

Moving the "actions" cell to the end seems like a good idea (seems an improvement regardless of a11y imo, now that I see it done), but:

1) I don't think I understand how it improves the screen-readability -- is it just the addition of `scope="row"`?
2) Wouldn't it be better (from either perspective) to have Title / ComposerArranger / Performer / PublisherLabel / Actions ?

src/components/Search.svelte
broadwell: Yeah, I had thought that the attribute would disappear from the DOM when set to `false`, but probably that was just seeing what I wanted to see... Anyway your fix makes it all work properly.

simonwiles: It's a curious thing, because HTML attributes are always text, so `attrib="false"` evaluates to `true` and it gets real messy.  Svelte 5 interprets setting the attribute to the empty string differently from Svelte 4 as well, I noticed* recently.

\* spent a while chasing...

src/components/Search.svelte
broadwell: Yeah the font comment was a red herring... This starts to feel like leveling a chair by cutting a little bit from one leg, but if it keeps those ellipses from appearing in my browser, I'll be happy.

src/components/Search.svelte
broadwell: Sure, `gap-y` would help... It's interesting that the 200% zoom requirement kind of necessitates building in the equivalent of a mobile/tablet-friendly responsive layout, even if you'd rather not -- in which case it's definitely preferable to have that in mind at the start of development, rather than the end.

broadwell:
Hmm, my assumption was just that the `scope="row"` for the `<th>` element would signal the screen reader to do... something differently when it reached that cell.
Anyway yeah the `title` cell is a good candidate for the row header as well. I suppose the Publisher/Label has a somewhat higher chance of being unique, but that's not actually a requirement for the row header (nor is it possible with this table). I'd be fine with changing to the order you suggest. It's maybe worth noting however that power users do know and search on the publisher and row numbers, even if the latter is just an arbitrary number to the uninitiated. Probably that's what motivated Craig to use the Label as the leftmost cell on the SUPRA roll table.

PR #217, opened 2024-05-22
Move the sidebar nav's position indicator widget to the proper location when the window is resized

broadwell:
This should solve the problem wherein the position indicator on the sidebar navigator mini-roll image doesn't update when the browser window is resized vertically. Such a resize causes the navigator's mini-roll image correspondingly to grow taller or shorter, so if the position indicator's, well, position doesn't change accordingly, it appears in the wrong place on the roll after the window resize.


simonwiles (approved):
Yeah, this is a good catch and a good fix, and I can confirm both the issue and the resolution.  Thanks!

PR #218, opened 2024-05-23
Tweaks

simonwiles:
Some more tweaks; all subject to further discussion, if desired.

### DIFF
--- a/src/components/PlaybackControls.svelte
+++ b/src/components/PlaybackControls.svelte
@@ -1,37 +1,34 @@
 <style lang="scss">
   #perform-controls {
+    font-family: SourceSans3, sans-serif;
+
     :global(kbd) {
       bottom: 0.35rem;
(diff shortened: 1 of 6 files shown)


broadwell:
I like everything so far -- the transport button hover styling is certainly an improvement, and bringing the `<ViewerMetrics>` "forward" in the layout really helps.
I agree that it's probably preferable for the pedal buttons to have the primary color styling, in which case the `<ViewerMetrics>` probably could just have the grey styling that's current applied to the buttons.
And I also concur that the pedal buttons need a hover style; the only stipulation I'd add is that it should be more subtle than their actual pressed/active style, because it's possible for each button to be either pressed or not pressed when hovered, and it's important to be able to ascertain its pressed status both when it's hovered and not.

broadwell (approved):
This all looks great to me -- thanks!

PR #219, opened 2024-06-11
Expanded page templates and content for user help documentation

broadwell:
As one of the commits indicates, the layouts, styles, text, images, page names and structuring of this help material is in pretty rough draft form and is highly likely to change. But rather than designating this as a draft/WIP for the foreseeable future, it's probably better to PR and merge it officially, so that the currently available content may prove helpful to those who visit the develop preview site, and also might prompt suggestions of desirable documentation content from other project stakeholders.


src/Pianolatron.svelte
```diff
@@ -378,6 +379,33 @@
...
+    if ($playbackProgressEnd < 1) {
+      params.end = ($playbackProgressEnd * 100).toFixed(2);
+    }
+    urlToCopy.search = new URLSearchParams(params);
+
+    try {
+      window.navigator.clipboard.writeText(urlToCopy.toString());
+    } catch {
+      // ignore error
+    }
```
simonwiles, line 402: Is this an appropriate place for a notification?  (In fact, in testing the functionality I think perhaps a notification on success would be equally valuable -- if you agree I'll add something in.)

simonwiles: Incidentally, do you remember the reason you changed the `&times;` for an `X` as the notification close button a few weeks ago?  Presumably it wasn't aesthetics?

simonwiles (approved):
Great, I think getting this in now is appropriate.  There are a bunch of things that will be improved here, of course, but this is (as has been proven) a very heavy wheel, so thanks doing the hardest part of getting it rolling.  Further bits and pieces will be easier for a bunch of reasons.

src/Pianolatron.svelte
broadwell: Yeah actually a notification on success is probably warranted, because the brief animation on the bookmark button is pretty easy to miss -- maybe just text that reads "Bookmark link copied to clipboard."
I wasn't sure what to do on failure (and the linter complains about a `console.log()`, for good reason), but I'd be in favor of a notification to the effect of "Unable to copy bookmark link to clipboard."

broadwell: And yes I think I changed `&times;` to `X` solely because the `&times;` was a bit small. There certainly are better ways to address that complaint, however...

simonwiles: Cool -- this is all fine with me.  I'll add the notifications.  (Tbh I wasn't expecting a button called "Bookmark" to copy something to my clipboard, so a notification that that's what's happened is a good idea imo.)

PR #220, opened 2024-06-13
Miscellaneous small tweaks (mostly for the docs pages)

simonwiles:
A couple of these are a bit opinionated; feel free to push back or reject.


broadwell:
Streamlining and rationalization of the doc layouts and file locations is welcome! I do have a few gripes with the new changes though, and for some reason pretty much all of these things look worse in Firefox on Mac than Chrome. For now I'll just screenshot my complaints below and give you the first pass at deciding how (and whether) to address them.

1. Can we get rid of this shadow below the header in the Listen view?
Screen Shot 2024-06-13 at 7 34 53 PM

2. The vertical alignment of the navbar links is noticeably more off-centerline towards the top of the bar in Firefox on Mac than Chrome; can we push those texts down by a few pixels?
Screen Shot 2024-06-13 at 7 35 13 PM

3. This just seems to me like an awful lot of blank space above the first doc header/content. Also, I'm not a big fan of the enlarged doc font size -- the contrast between the documentation text and the other font sizes in the app is kind of jarring. Assuming we won't need to bump up all of the app's font sizes for a11y reasons (perhaps a naive assumption), could we maybe choose a middle ground between the doc font sizes in the PR and the previous setting?
Screen Shot 2024-06-13 at 7 35 27 PM

5. The controller graphic shrinks to fit the width of the main content column on Chrome but just overflows on Firefox for Mac.
Screen Shot 2024-06-13 at 7 35 45 PM

simonwiles:
> 1. Can we get rid of this shadow below the header in the Listen view?

5e618cea reverts b3892c4f.

> 2. The vertical alignment of the navbar links is noticeably more off-centerline towards the top of the bar in Firefox on Mac than Chrome; can we push those texts down by a few pixels?

> 4. The controller graphic shrinks to fit the width of the main content column on Chrome but just overflows on Firefox for Mac.

I can't reproduce either of these here (looks identical on Chromium and FF), so I'm not immediately sure how to address them.  Will fire up a Mac to have a look.

> 3. This just seems to me like an awful lot of blank space above the first doc header/content. Also, I'm not a big fan of the enlarged doc font size -- the contrast between the documentation text and the other font sizes in the app is kind of jarring. Assuming we won't need to bump up all of the app's font sizes for a11y reasons (perhaps a naive assumption), could we maybe choose a middle ground between the doc font sizes in the PR and the previous setting?

Yeah, fair enough, I guess we just have a difference of opinion here;  I would go larger and more whitespace, if anything.  2bceb708 reverts 8aec65bc and a337104f and puts things back as they were.

simonwiles:
> I can't reproduce either of these here (looks identical on Chromium and FF), so I'm not immediately sure how to address them. Will fire up a Mac to have a look.

So I can see (2) on macOs (both FF and Chrome look off compared to on Linux, where the text is perfectly aligned on both FF and Chromium, but FF is worse).  The images look fine on FF on Catalina so I borrowed a newer Mac and I can see what you're seeing with FF on Sonoma.  Sigh.

Since I've got two macbooks open in front of me now, I'll come up with something.

Edit: this is rather odd; there's a whole section of CSS declarations that's being ignored on FF on macOs on Sonoma -- looks like a nesting bug.  See also the link color that's not being applied on the credits page.

I've just noticed I have FF 127.0 on Catalina here, but 115.12 ESR on the Sonoma system.  This could be relevant.

simonwiles:
Okay, I think I've address all those issues; apologies for the saga :)

broadwell (approved):
All right, great, thanks for humoring my nitpicks... Everything looks fine in FF and Chrome on Mac OS now. I wouldn't be averse to bumping up the body text size on the docs pages, but maybe we can revisit that question if/when we get feedback on it from other potential users.

PR #221, opened 2024-06-18
Export MIDI reflecting current in-app expression settings; numerous other changes

broadwell:
There's no particular rush to merge this -- the PR is partly just to get a convenient Netlify preview link.

The main contribution here is the in-app expression MIDI export functionality, though the PR also now includes several more updates that are nearly as consequential. Some no doubt could be moved to other PRs, but perhaps that's not necessary. The commit messages should (hopefully) explain all of the changes.


src/components/RollDetails.svelte
```diff
@@ -106,7 +106,7 @@
     {@html metadata.publisher || unavailable}
   </dd>
   {#if similarWorksByPerformer.length > 0}
-    <dt>Similar Works By This Performer</dt>
+    <dt>Other Works By This Performer</dt>
```
moritztim: I see you chose the solution of changing the wording, rather than using a fancy matching algorythm to determine musical similarity :/

src/Pianolatron.svelte
```diff
@@ -247,6 +247,7 @@
     return Promise.all([mididataReady, metadataReady, pianoReady]).then(
       ([, metadataJson]) => {
         metadata = (({ holeData: _, ...obj }) => obj)(metadataJson);
+        metadata.druid = roll.druid;
```
moritztim, line 250: That's a much simpler solution, I'll close [my pr](https://github.com/sul-cidr/pianolatron/pull/222).

src/components/RollDetails.svelte
broadwell: Yes, there's certainly a lot more we could do in the way of recommendations via basic metadata indexing or fancy roll similarity metrics, but for now it's best to be honest!

simonwiles (approved):

PR #222, opened 2024-06-19
add missing metadata property druid

moritztim:
`RollDetails` attempts to access `metadata.druid`, which is always undefined. This leads to `Similar Works By This Performer` including the original work itself, because this check is essentially bypassed.
https://github.com/sul-cidr/pianolatron/blob/236a9c8f97fc6b57d4a885b83190c4b22c8775cf/src/components/RollDetails.svelte#L72-L74
the `druid` is not contained in the `metadataJson`, so it needs to be added after fetching it.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -235,7 +235,13 @@
 
     metadataReady = fetch(joinPath("json", `${roll.druid}.json`))
       .then((metadataResponse) => {
-        if (metadataResponse.status === 200) return metadataResponse.json();
+        if (metadataResponse.status === 200)
+          return metadataResponse.json().then(
+              (metadataJson) => {
+                metadataJson.druid = roll.druid;
+                return metadataJson;
+              },
+            );
         throw new Error("Error fetching metadata file! (Operation cancelled)");
       })
       .catch((err) => {


moritztim:
Take [`zb497jz4405`](https://deploy-preview-222--pianolatron-develop.netlify.app/?druid=zb497jz4405) as an example, it should only have one similar work.
In the [current version](https://deploy-preview-220--pianolatron-develop.netlify.app/?druid=zb497jz4405) there are two, one of them being `zb497jz4405` itself.

broadwell:
Thanks for this, and thanks in advance for reporting any other issues you might encounter!
Coincidentally, we noticed this bug earlier in the week, and there is a (slightly different) fix for it in [this commit](https://github.com/sul-cidr/pianolatron/pull/221/commits/8ab4b02580de5a71b8b6960b1c8114a16e148667) of https://github.com/sul-cidr/pianolatron/pull/221. So it should be addressed soon, one way or the other.

FYI most further changes we might make in the next couple of months *should* be fairly superficial and involve non-core features that are likely to be cut away from the exhibit version. But let us know if you'd like to discuss our roadmap and possible ways to synchronize our efforts.

moritztim:
> non-core features that are likely to be cut away from the exhibit version.

I will soon show my current version to the museum. I'll get back to you once I know exactly how the exhibit should look like and what features will be included.

> possible ways to synchronize our efforts.

Tbh, I've been kinda brute forcing changes that aren't exactly scalable or configurable, but I'll make pull requests for any cherry-pickable commits that might be beneficial to the project

moritztim:
https://github.com/sul-cidr/pianolatron/pull/221/files#r1647579718

PR #225, opened 2024-07-01
Further a11y work

simonwiles:


broadwell (approved):
This all works for me on multiple browsers, and I have no suggestions for refining the implementation. It also ticks all of the applicable boxes on my own scoresheet of SODA to-dos, so I'll go ahead and merge it.

PR #226, opened 2024-07-08
Miscellaneous Updates

simonwiles:
This is a couple of minor linting-type fixes, and a perhaps controversial one (efe081c1c94a931d19e5b9b1958e1f09588deb3b).  There was supposed to be more, but this has been languishing for a week and I've not added anything else to it yet, so I'm PRing now, and will do another PR later.


broadwell (changes requested):
You're right, everything looks great except perhaps https://github.com/sul-cidr/pianolatron/commit/efe081c1c94a931d19e5b9b1958e1f09588deb3b -- which is functionally an improvement, for sure, but it does kind of go against what I see as the main priority for the header behavior, which is just to minimize the degree of visual surprise when going back and forth between the SUPRA exhibit and the Pianolatron search and play modes (obviously the perform mode is kind of the odd duck in this regard).
Choosing either path at this point involves a lot of speculation about stakeholder preferences and the trajectory of the Spotlight integration, but I think I'd prefer to leave https://github.com/sul-cidr/pianolatron/commit/efe081c1c94a931d19e5b9b1958e1f09588deb3b out for now.
Also it's odd that the Netlify preview didn't build, but I can't see how that could be anything but a problem on their end.

simonwiles:
Okay, no problem, offending commit has been dropped 👍️.

PR #228, opened 2024-07-12
Subtle layout changes

broadwell:
Might as well PR this. Objections and alternatives (especially to 7b4d168741a5f028e2f9a491a2953c9db63b8f9a) are welcome.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -15,7 +15,7 @@
   }
 
   .listen-app {
-    height: calc(100vh - 220px);
+    height: calc(100vh - 217px);
   }
   .perform-app {
     height: calc(100vh - 170px);
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -51,6 +51,7 @@
       position: absolute;
       right: 0;
       top: 0;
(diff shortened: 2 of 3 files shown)


simonwiles (approved):
Happy to approve and merge this.

One observation (at the risk of taking us further down what I already regard to be something of a fool's errand of a rabbit hole 🤪), should the `<StanfordHeader/` component at the very top have "Stanford | Libraries" and the logo too?  The "Sign in" and "Feedback" links?  Where does this end?  We could iFrame the header in? 😬

broadwell:
Yes, following such a line of thought does tend to expose some logical contradictions at the core of this enterprise, which for various reasons has us pursuing (for now) a "name-brand knockoff" strategy, i.e., it should look identical at a casual glance, whereas differences become apparent upon closer inspection.

iFraming in parts of the header is probably the only sustainable way to get certain desirable similarities to work, most notably the dropdown content under "Roll Archive Features" (though it's also likely that the contents of the Spotlight page's menu items will change in the near future -- removing the Search widget being the most probable). Additionally, stakeholders have also indicated that it's preferable for parts of the header stack to be customized for the Pianolatron, specifically that the top `<StanfordHeader/` should be more organizationally inclusive than the Libraries header. So 🤷

PR #229, opened 2024-07-16
Searchbox Fixes

simonwiles:
This PR fixes the bug when a single punctuation character is entered (but please double check on Safari!), and another related to highlighting going wrong due to irregular spacing in the labels from the upstream metadata.

With respect to the initial impetus regarding the odd behaviour searching for, e.g. "stars and stripes" it wasn't so much a bug (the code was working as written) as a peculiar choice/oversight on my part when (re-)implementing the search box on the search page.  Behaviour is much better and less wtf now :)

### DIFF
--- a/src/components/Search.svelte
+++ b/src/components/Search.svelte
@@ -371,7 +371,7 @@
 
     if (searchParts.length) {
       filteredListItems = filteredListItems.filter((item) =>
-        searchParts.some((searchPart) =>
+        searchParts.every((searchPart) =>
           searchFields.some((field) =>
             normalizeText(item[field]).includes(searchPart),
           ),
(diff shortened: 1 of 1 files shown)


broadwell (changes requested):
This all looks great, especially switching to AND logic, but can you think of a minimally invasive fix to make the multi-line matches appear on separate lines (or alternatively with something like a `/` between them)? This is really only an issue with the `Composer / Arranger` field and is presumably related to the blank-space cleanup in https://github.com/sul-cidr/pianolatron/pull/229/commits/880128c844a7193dc9e47e2ec9c0db2c2492192c.
Screen Shot 2024-07-16 at 11 13 09 AM

simonwiles:
Ahh, good catch, thanks; committed a fix.

broadwell (approved):
Great! Yes this all looks to me now.

PR #230, opened 2024-07-16
Update roll data due to upstream parsing changes; streamline treatment of 65-note roll(s)

broadwell:
The MIDI and JSON data changes primarily derive from running the updated `roll-image-parser` code (which was already applied to the new and existing 88-note rolls) to the red Welte rolls as well. The main update involved finessing how end-of-track messages are written to the treble and bass tracks in the expression MIDI files; the previous method produced an occasional bug in Expression MIDI mode that would cause some or all notes in the final grand cadence to be rather anticlimactically truncated -- `rk796xd9697` was a prime example.

The other changes streamline the handling of our lone 65-note roll a bit more and indicate visually when certain roll emulation settings are disallowed by the roll type. 65-note rolls get no expression emulation treatment on the backend (via `midi2exp`) or in the app, which kind of makes sense given that they have no accents or pedaling, but this also means that low-level emulation factors like acceleration and tracker hole extension also aren't applied. I briefly considered modifying `midi2exp` and/or adding a minimal (by design) `SixtyFiveNoteExpressionizer` to the app to do this, but considering 65ers' rarity and overall fairly low interest to researchers, I think it's preferable just to continue treating our singe roll as an exception until more appear in the pipeline.


simonwiles (approved):
LGTM -- some checkbox-related suggestions to follow in a subsequent PR.

PR #231, opened 2024-07-16
UI for when roll emulation settings are not available

simonwiles:
This is marginal gains for a rare edge case, but unobjectionable I hope.  The `:has` selector has pretty good support these days, but the use here will just degrade acceptably where it's not available.

### DIFF
--- a/src/components/AdvancedSettings.svelte
+++ b/src/components/AdvancedSettings.svelte
@@ -162,7 +162,8 @@
       Play Expressions:
       <input
         type="checkbox"
-        bind:checked={$playExpressionsOnOff}
+        checked={$rollHasExpressions && $playExpressionsOnOff}
+        on:change={() => ($playExpressionsOnOff = !$playExpressionsOnOff)}
         disabled={!$rollHasExpressions}
       />
     </div>
@@ -170,7 +171,8 @@
       Use Roll Pedaling:
       <input
         type="checkbox"
-        bind:checked={$rollPedalingOnOff}
+        checked={$rollHasExpressions && $rollPedalingOnOff}
+        on:change={() => ($rollPedalingOnOff = !$rollPedalingOnOff)}
         disabled={!$rollHasExpressions}
       />
     </div>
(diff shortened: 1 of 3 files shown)


broadwell:
Yeah apparently `:has` is "coming soon" to an official Firefox release: https://www.mozilla.org/en-US/firefox/121.0beta/releasenotes/

simonwiles:
> Yeah apparently `:has` is "coming soon" to an official Firefox release: https://www.mozilla.org/en-US/firefox/121.0beta/releasenotes/

? I'm using Firefox 128.0 here...

[`:has` support is decent](https://caniuse.com/css-has) -- I'd hesitate to use it for anything load-bearing, but for progressive enhancements / nice-to-haves, I reckon it's green lights now.

broadwell:
I'm inclined to merge this now, because everything works as intended (and also to get the next a11y PR in), but I'd like for one of the next PRs to involve also converting the checkbox items under the MIDI Settings and Expression Settings menus to `<fieldset><label><input>` from `<fieldset><div><input>` so that they have the same behavior, i.e., the checkbox assumes the highlight color when the box OR the label is hovered (assuming that's desirable) and toggles when either the box or the label is clicked.

simonwiles:
> I'd like for one of the next PRs to involve also converting the checkbox items under the MIDI Settings and Expression Settings menus to `<fieldset><label><input>` from `<fieldset><div><input>` so that they have the same behavior

Yeah, that's a good point, this has introduced an inconsistency there.  It's not exactly like the UI is otherwise free of inconsistencies, of course, but we want to trend in a positive direction, not the other way around, sorry :)  I'll take care of that tomorrow 👍️.

broadwell:
> ? I'm using Firefox 128.0 here...

Turns out my FF install was moved to the ESR track when the OS aged out, and I guess it doesn't automagically switch back to the non-ESR update channel upon a subsequent OS update. Time to reinstall...

simonwiles:
Huh, that's a bit too clever for my taste, but I suppose it makes some sense.  Coincidentally (ironically?) FF 128, just released, is the new ESR version...

broadwell:
> Huh, that's a bit too clever for my taste, but I suppose it makes some sense. Coincidentally (ironically?) FF 128, just released, is the new ESR version...

Yeah it's available, but my FF Mac OS install was in no hurry to leave 115.13.0esr. Anyway I've reinstalled from the main channel and now have full use of `:has`.

PR #233, opened 2024-07-18
Minor hanges

simonwiles:
Very minor changes, but better now than later.

### DIFF
--- a/src/components/ViewerMetrics.svelte
+++ b/src/components/ViewerMetrics.svelte
@@ -9,10 +9,15 @@
     display: inline-block;
 
     .metric {
+      align-items: center;
       display: flex;
       min-height: 1.25em;
       justify-content: space-between;
       position: relative;
+
+      span:last-child {
+        font-family: monospace;
+      }
     }
   }
 </style>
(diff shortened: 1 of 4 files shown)


src/components/ViewerMetrics.svelte
```diff
@@ -13,6 +13,10 @@
       min-height: 1.25em;
       justify-content: space-between;
       position: relative;
+
+      span:last-child {
+        font-family: monospace;
+      }
```
broadwell, line 20: Yes the monospace readout is an improvement, but I find it a bit distressing that the baselines (or even centerlines) of the labels and readouts are noticeably (to me) out of alignment horizontally.
Screen Shot 2024-07-18 at 12 47 22 PM
I think maybe adding an `align-items: baseline` or `align-items: center` to the `.metric` class would be an improvement? Let me know what you think.
Screen Shot 2024-07-18 at 12 56 35 PM

simonwiles: Yeah, it's a lot more egregious on some browsers than others, and I'd not picked it up.  I prefer `align-items: center` -- so I've added that to the last commit and force-pushed.  Thanks! 👍️

PR #234, opened 2024-08-15
Search page without headers and footer

broadwell:
Just PRing this to get the preview link, to be used for `iframe`ing experiments.


broadwell:
Superseded by #235 (which is still a draft).

PR #236, opened 2024-10-16
Avoid unnecessary wear and tear on Disklavier-type digital player pianos

broadwell:
See commit messages. The changes touch a small corner of the code/functionality (I hope), so we might as well merge (if it passes review), even though it's difficult to test without getting access to a Disklavier.

### DIFF
--- a/src/components/SamplePlayer.svelte
+++ b/src/components/SamplePlayer.svelte
@@ -147,6 +147,8 @@
     midiSamplePlayer.pause();
     midiSamplePlayer.triggerPlayerEvent("pause");
     stopAllNotes();
+    webMidi?.sendMidiMsg("CONTROLLER", "SUSTAIN", false);
+    webMidi?.sendMidiMsg("CONTROLLER", "SOFT", false);
     $isPlaying = false;
   };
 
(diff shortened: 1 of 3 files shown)


simonwiles:
All good, thanks!

W/r/t the visual discrepancy, we already visually release the keys when paused, and I think it would be fine to release the pedals too, tbh?  I'd be more concerned that the pedalling status would be incorrect when playback is resumed (in the case where one or both pedals were active when playback is paused), which is something we've thought about before, I recall.  Perhaps a case could be made for preserving state out-of-band and then releasing everything when playback is paused, and restoring before playback resumes.

Happy to merge this now in any event.

simonwiles (commented):
👍️

broadwell:
Thanks for the merge! Also, re:

>Perhaps a case could be made for preserving state out-of-band and then releasing everything when playback is paused, and restoring before playback resumes.

I think this would make good sense in the case that the playback is just paused and then resumed without moving the playback point -- though only for the pedaling state\*. But if the play point is moved at all during the pause, the analogous behavior of an actual reproducing piano becomes __undefined__, and it's probably more "realistic" just to reset the state as best we can and then have the system respond to the next control holes it sees after playback resumes. So rather than differentiate between these two situations, I'm OK with the Disklavier always conforming to the second paradigm.

\* One could presumably induce a Disklavier to restore the keypress state without actually restriking the notes (I think restrikes would be very unwelcome) by sending MIDI events with very low velocities, but this would only have a visual effect and is almost certainly not worth the hassle.

PR #238, opened 2024-12-14
Place active/mouseover hole tooltips below on downward-scrolling rolls

broadwell:
This hole overlay display affordance is becoming more important as we add more downward-scrolling roll types (which is everything besides Red Welte). I figured I'd PR it on its own in case you have a better way to do it.

### DIFF
--- a/src/styles/hole-highlighting.scss
+++ b/src/styles/hole-highlighting.scss
@@ -23,6 +23,8 @@ $highlight-hover-outline-offset: 4px;
       }
 
       // the label "flag" (points to the right by default -- see .flag-left below)
+      // (also is positioned at the top of the label for upward-scrolling rolls,
+      // at the bottom for downward-scrolling rolls -- see .flag-bottom below)
       &::after {
         background-color: $highlight-hover-outline-color;
         color: white;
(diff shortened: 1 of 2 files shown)


src/styles/hole-highlighting.scss
```diff
@@ -102,6 +117,12 @@ $highlight-hover-outline-offset: 4px;
       );
       padding: 8px 4px 8px ($highlight-hover-outline-width + 4px);
     }
+
+    // and .flag-bottom marks
+    mark.flag-bottom:hover::after {
+      top: auto;
+      bottom: -($highlight-hover-outline-offset + $highlight-hover-outline-width);
+    }
```
simonwiles: For the case where "Show Details for Active Notes" is enabled _*and*_ the mark is hovered:
image

the `transform` needs to be wound off again.  I'll add a commit to this branch, and wait for you to okay it before merging.

simonwiles (approved):
Yep, this looks sound to me.  Something fancy that keeps the flag in the viewport would be nice, but a bit complicated (would require a bunch of JS, I think?), and this is pretty good (good enough, I think).

broadwell:
Thanks for the review, and for the fix! This is especially helpful with Green Welte and Duo-Art rolls, which can have very long control perforations.
I suppose it would be ideal if the tooltips would stay visible even if both ends of the perforation are outside the viewer window, but as you note that probably would be more trouble than it's worth to implement.

PR #240, opened 2026-02-24
Accessibility Review 2026

simonwiles:
Includes some dependency updates and deprecation mitigation.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -13,6 +13,11 @@ export default defineConfig({
       preprocess: [
         preprocess.scss({
           prependData: `@use "sass:math";@use 'src/styles/sass-globals.scss' as *;`,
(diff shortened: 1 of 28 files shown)


broadwell (approved):
This is all excellent work. The main functionality I was able to test were the keyboard (tab) navigation and interaction features, which worked quite well on a variety of browsers on Mac OS and Windows, and the voiceover navigation support, which was also fine.
The only surprise was that the graphical pedals are selectable via keyboard navigation in Perform mode, but can't be activated ("clicked") via the space or enter keys. Is this the intended behavior, and/or the most reasonable to implement? It might be somewhat surprising to users, but not a show-stopper, especially since there are at least two other ways of activating those pedals via the keyboard in Perform mode.

PR #241, opened 2026-03-31
Better layout on resize and when keyboard maximized

broadwell:
This is a suggestion for how to change the [more-accessibility](https://github.com/sul-cidr/pianolatron/tree/more-accessibility) branch (specifically commit [e627ac2](https://github.com/sul-cidr/pianolatron/commit/e627ac206dbdfcc1389f3490f4196876a7a571f9)) so that the sidebars still are no longer collapsible, but the gridded layout behavior of the sidebar(s), roll viewer and piano keyboard (if visible) is the same as previously.

The layout shortcomings of the current version of [more-accessibility](https://github.com/sul-cidr/pianolatron/tree/more-accessibility) are evident when the page is resized vertically, especially when the piano keyboard is maximized.
Screenshot 2026-03-31 at 1 48 24 AM

Note that the previous (more desirable) layout behavior still does have its quirks, such as how the left sidebar can y-overflow in listen mode when it really has vertical room to spare -- this apparently (?) is so that space will be available if the keyboard is maximized. It would be nice to improve this, but it need not be a high priority.
Screenshot 2026-03-31 at 1 43 11 AM

Anyway, we probably can just delete the `<FlexCollapsible>` component entirely if this fix is sufficient?

Note also that the `height: calc(100vh - 135px);` is an unrelated change to ensure that the app window doesn't have a vertical scroll bar in listen mode -- the previous value (`124px`) is a leftover from a different page header size.

### DIFF
--- a/src/Pianolatron.svelte
+++ b/src/Pianolatron.svelte
@@ -27,7 +27,7 @@
   }
 
   .listen-app {
-    height: calc(100vh - 124px);
+    height: calc(100vh - 135px);
   }
   .perform-app {
     height: 100vh;
@@ -38,11 +38,6 @@
 
   :global(#left-sidebar) {
     grid-area: left;
-
-    p {
-      opacity: 0.5;
-      padding: 0.5em 1em;
-    }
   }
 
   #roll {
@@ -70,6 +65,7 @@
   }
 
   .sidebar {
+    position: absolute;
     display: flex;
     flex-direction: column;
     gap: 0.5em;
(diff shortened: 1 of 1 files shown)


simonwiles (approved):

simonwiles:
Thanks -- this is all fine with me.  The unneeded import is stripped at build time, but yeah, the component file can probably go now; I'll add that to the parent branch.

I can't reproduce the issue in your second screenshot, with FF or Chromium, with or without this change...

broadwell:
Great! And yeah, maybe the issue in the second screenshot is actually related to how scrollbars are (or aren't) rendered in Mac OS or something...

PR #244, opened 2026-04-20
Miscellaneous UX refinements

broadwell:
This PR includes tweaks to the search page, some minor text updates to the documentation and menus, modifications to the roll panning behavior, significant refactoring of how the RollViewer loads and visualizes in-app expression data, and the return of the collapsible sidebars.


src/components/Search.svelte
```diff
@@ -619,6 +619,14 @@
       >
         Next &raquo;
       </button>
+      <button
+        disabled={currentPage * pageSize >= filteredListItems.length}
+        on:click={() => {
+          currentPage = filteredListItems.length / pageSize;
+        }}
```
simonwiles, line 626: This can (and does) result in a `currentPage` value that isn't an integer, which causes odd behaviour if paging back far enough (try hitting "Last" and then paging back with "Previous" until you get to the beginning!).

I think the best behaviour is to skip to the actual last page (even though that won't generally be a full one).  I'll add a commit.

src/components/SamplePlayer.svelte
simonwiles, line 1: These changes are commit `22c03c75`, "Remove redundant reactive statements", but they're actually just changes made by the linter/prettifier on save; unclear if these are supposed to be removed or not?

simonwiles (approved):
Thanks for the detailed and organized commits -- very easy to follow.  I like all these changes :)  For the "Disable animation of vertical pan on roll drag" ([bd76da1](https://github.com/sul-cidr/pianolatron/pull/244/commits/bd76da16ebff003bea9d02e6554562ca4e33deba)) I think I may have an alternative fix, but I agree with your reasoning and think your new behaviour is probably better even if the previous behaviour could be "fixed".

Going to add a commit as mentioned in the comment below, and then I'll go ahead and merge this.  Thanks!

src/components/SamplePlayer.svelte
broadwell: Oops - yeah, the actual redundant reactive statements were in `RollViewer.svelte` around line 1050 and were removed as part of the previous commit (https://github.com/sul-cidr/pianolatron/pull/244/commits/bd76da16ebff003bea9d02e6554562ca4e33deba). Apparently I saw the reordering of the statements here in `SamplePlayer.svelte` that was done by format-on-save and mistakenly assumed this was where I had made the aforementioned redaction. Well, it should be fine to leave this commit in even though it has no actual effect on anything.

src/components/Search.svelte
broadwell: Right, I definitely didn't think this through or test it quite enough, so the fix (and the styling update) are much appreciated!

broadwell:
Thanks! Yeah https://github.com/sul-cidr/pianolatron/pull/244/commits/bd76da16ebff003bea9d02e6554562ca4e33deba weaves kind of a tangled web to get to a solution, but if you can live with it then I think this is good to go.

src/components/SamplePlayer.svelte
simonwiles: Yeah, no worries -- I nearly just renamed the commit to "Linting" and moved on, but I just wanted to check that there wasn't supposed to be anything substantive here apart from the defensive wrapping of the expressions in parentheses by the auto-formatter 👍️

PR #245, opened 2026-04-21
Performance improvements

simonwiles:
Basically two (related) interventions here:

2f67eb7983c279e2747794dddd63bb19163123f0 creates a `$throttledTick` store (for use in the visualization only) which batches updates according to the animation frame rate, to avoid doing more work than necessary (`$currentTick` can, in theory, update up to 1,000 times per second, and does generally in practice exceed 500 times per second).

fb65da8803f302cd2432a7c40edf033f2dcbbd21 replaces the (now throttled) DOM-based hole highlighting with an approach based on a `<canvas/>` element.

Together they seem to make a noticeable improvement, especially on Firefox and on less powerful devices.  I _think_ the user-facing behaviour is otherwise entirely unchanged, but there are a bunch of edge cases that I might not have tested thoroughly.  However, I think the perf wins are worth the churn and the risk of an edge case slipping through and needing to be fixed later.  Lmk if you agree :)

### DIFF
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -146,7 +146,7 @@
     useInAppExpression,
     userSettings,
   } from "../stores";
-  import { clamp, getHoleLabel } from "../lib/utils";
+  import { clamp, defaultHoleColor, getHoleLabel } from "../lib/utils";
   import RollViewerControls from "./RollViewerControls.svelte";
   import RollViewerScaleBar from "./RollViewerScaleBar.svelte";
   import AriaAnnouncer from "../ui-components/AriaAnnouncer.svelte";
@@ -752,10 +752,11 @@
       // Get highlight color
       let color = hole.color;
       if (
-        !$userSettings.showNoteVelocities ||
-        $userSettings.highlightEnabledHoles
+        hole.type === "note" &&
+        (!$userSettings.showNoteVelocities ||
+          $userSettings.highlightEnabledHoles)
       ) {
-        color = "60, 100%, 50%"; // yellow default for non-velocity mode
+        color = defaultHoleColor; // yellow default for non-velocity mode
       }
       if (
         !$rollPedalingOnOff &&
@@ -766,7 +767,7 @@
       }
 
       // Animate highlight: bright "attack" then fade to steady opacity over 500ms
-      let opacity = 0.5;
+      let opacity = 0.6;
       let glow = 8;
       if (!_highlightActivation.has(hole)) {
         _highlightActivation.set(hole, performance.now());
@@ -775,7 +776,7 @@
       if (elapsed < 500) {
         const t = elapsed / 500;
         const eased = easeInOutCubic(t);
-        opacity = 1 - 0.5 * eased;
+        opacity = 1 - 0.4 * eased;
         glow = 8 * (1 - eased);
       }
 
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -124,7 +124,6 @@
   import {
     avgHoleWidth,
     bassExpCurve,
-    currentTick,
     drawVelocityCurves,
     expressionParameters,
     firstHolePx,
@@ -1193,14 +1192,14 @@
     }
     updateSelectionOverlays();
     updateVisibleSvgPartitions();
-    updateViewportFromTick($currentTick);
+    updateViewportFromTick($throttledTick);
   };
 
   /* eslint-disable no-unused-expressions, no-sequences */
   $: ($playbackProgressStart, updateSelection());
   $: ($playbackProgressEnd, updateSelection());
-  $: updateViewportFromTick($currentTick);
-  $: ($transposeHalfStep, rehighlightHoles($currentTick));
+  $: updateViewportFromTick($throttledTick);
+  $: ($transposeHalfStep, rehighlightHoles($throttledTick));
   $: ($drawVelocityCurves,
     partitionExpressionOverlaySvgs($bassExpCurve, $trebleExpCurve));
   $: ($userSettings.activeNoteDetails,
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -772,12 +772,20 @@
       highlightCtx.fill();
 
       // Draw detail label on canvas when active-note-details is enabled
-      if ($userSettings.activeNoteDetails && hole.type === "note") {
-        const noteLabel = hole.label;
+      if ($userSettings.activeNoteDetails) {
+        let noteLabel = hole.label;
         let velocityLine = "";
-        if ($userSettings.showNoteVelocities) {
-          const vel = $playExpressionsOnOff ? (hole.v ?? 64) : 64;
-          velocityLine = `v:${Math.round(vel)}`;
+
+        if (hole.type === "note") {
+          noteLabel = getHoleLabel(
+            hole.m + $transposeHalfStep,
+            $rollMetadata.ROLL_TYPE,
+          );
+
+          if ($userSettings.showNoteVelocities) {
+            const vel = $playExpressionsOnOff ? (hole.v ?? 64) : 64;
+            velocityLine = `v:${Math.round(vel)}`;
+          }
         }
 
         highlightCtx.save();
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -772,7 +772,65 @@
       const radius = Math.min(6, screenW / 4, screenH / 4);
       highlightCtx.roundRect(screenX, screenY, screenW, screenH, radius);
       highlightCtx.fill();
+
+      // Draw detail label on canvas when active-note-details is enabled
+      if ($userSettings.activeNoteDetails && hole.type === "note") {
+        const noteLabel = hole.label;
+        let velocityLine = "";
+        if ($userSettings.showNoteVelocities) {
+          const vel = $playExpressionsOnOff ? (hole.v ?? 64) : 64;
+          velocityLine = `v:${Math.round(vel)}`;
+        }
+
+        highlightCtx.save();
+        highlightCtx.textAlign = "center";
+
+        const cx = screenX + screenW / 2;
+        if (!$scrollDownwards) {
+          drawTextLine(
+            highlightCtx,
+            noteLabel,
+            true,
+            cx,
+            screenY + screenH + 36,
+          );
+          if (velocityLine) {
+            drawTextLine(
+              highlightCtx,
+              velocityLine,
+              false,
+              cx,
+              screenY + screenH + 58,
+            );
+          }
+        } else {
+          if (velocityLine) {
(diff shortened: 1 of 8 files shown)


src/components/RollViewer.svelte
```diff
@@ -1051,11 +1130,20 @@
   $: ($playbackProgressStart, updateSelection());
   $: ($playbackProgressEnd, updateSelection());
   $: updateViewportFromTick($currentTick);
```
broadwell: Consider tying reactive calls to `updateViewportFromTick` to `$throttledTick`, as is done with the highlight redraws -- this would sync the scrolling of the roll image tiles to the visual refresh rate (I think), which seems OK, and presumably would offer some kind of efficiency savings?

src/components/RollViewer.svelte
broadwell, line 1269: I'm no longer seeing the active note tooltips when the notes are played and "Show Details for Active Notes" is selected in the advanced Visualization Settings.
Screenshot 2026-04-21 at 11 22 55 PM

broadwell (changes requested):
These look like great additions! One suggestion (the full implications of which I admittedly haven't thought through) and two bug reports are provided in the comments.

src/components/RollViewer.svelte
```diff
@@ -1108,6 +1196,8 @@
     {/if}
   {/if}
 
+  <canvas id="active-note-highlight-canvas" bind:this={highlightCanvas} />
```
broadwell, line 1292: I'm also getting some undesirable overlay behavior when I pause playback with active notes highlighted and pan or scroll the roll.
(This probably isn't the relevant code snippet exactly, but I've established a precedent for reporting bugs in code comments, so I guess I'll just continue to do so.)

Image

src/components/RollViewer.svelte
simonwiles: Yeah, I went through a couple of cycles of variations on this in earlier iterations of this commit.  I found it was creating other issues (broke the rewind, caused the percentage marker in the navigator to get out-of-sync, a dew other things) and decided there wasn't enough performance gain to make it worth it.  However, I changed a few things since then, and just naively swapping `$currentTick` for `$throttledTick` here doesn't break as many things as it did before, so it may be worth revisiting, thanks :+1:  Will update when I've done some more work.

src/components/RollViewer.svelte
simonwiles: Yeah this is related to a whole class of bugs I missed (e.g. when zooming in and out, too), thanks, mea culpa.  Fix incoming.

src/components/RollViewer.svelte
```diff
@@ -772,9 +772,67 @@
       const radius = Math.min(6, screenW / 4, screenH / 4);
       highlightCtx.roundRect(screenX, screenY, screenW, screenH, radius);
       highlightCtx.fill();
+
+      // Draw detail label on canvas when active-note-details is enabled
+      if ($userSettings.activeNoteDetails && hole.type === "note") {
```
broadwell: We also need code here to display the label if it's a control hole, and to redraw the label when the transposition is changed while the note is still under the tracker bar. I'd suggest something like

Replaces lines 777-783:
```
      if ($userSettings.activeNoteDetails) {
        let noteLabel = hole.label;
        let velocityLine = "";

        if (hole.type === "note") {
          noteLabel = getHoleLabel(
            hole.m + $transposeHalfStep,
            $rollMetadata.ROLL_TYPE,
          );

          if ($userSettings.showNoteVelocities) {
            const vel = $playExpressionsOnOff ? (hole.v ?? 64) : 64;
            velocityLine = `v:${Math.round(vel)}`;
          }
        }
```

src/components/RollViewer.svelte
simonwiles: It turns out that the implementation of `throttleTick` that I ended up committing didn't cause too many problems with `updateViewportFromTick` -- just needed to update `resetPlayback` and the function that loads expression settings.  Can't detect a noticeable performance improvement from this change, but it shouldn't hurt and now `<RollViewer/>` depends on `throttledTick` only throughout, which feels much better.  50765d20 and a703703a.

src/components/RollViewer.svelte
simonwiles: Resolved in 871531eb.

src/components/RollViewer.svelte
simonwiles: Resolved in 04775906.

src/components/RollViewer.svelte
simonwiles: Yeah, that's neat, thanks :+1:  84077402

src/components/RollViewer.svelte
```diff
@@ -772,9 +772,67 @@
...
+        if ($userSettings.showNoteVelocities) {
+          const vel = $playExpressionsOnOff ? (hole.v ?? 64) : 64;
+          velocityLine = `v:${Math.round(vel)}`;
+        }
+
+        highlightCtx.save();
+        highlightCtx.textAlign = "center";
+
+        const cx = screenX + screenW / 2;
+        if (!$scrollDownwards) {
```
broadwell, line 818: **Edited:** the text overlay positions (and the presence of the backing screen) I'm seeing on Mac OS and Windows are actually, and unfortunately, quite different from what I get on Linux, with a similarly high-res monitor, so the below may need some further consideration...

**Previous comment:**
I think the dark backing fill, bright text and somewhat larger font than previously are all desirable for accessibility reasons, even though the larger control hole labels do tend to get cropped by the viewer unless the user zooms out a little.

I'm noticing some misalignments of the texts and backing fills, however, on all of the OSes and monitors I've tried (Mac OS, ~~Linux~~ and Windows, all in 4K or otherwise "retina"). Hopefully the following suggested tweaks to the magic layout numbers are broadly applicable (**edit:** nope, not for Linux) ~~and we're not at the mercy of dynamic font scaling or something. I'd also just prefer to move the annotations some what closer to the holes than they have been previously:~~

```
const drawActiveHighlights ...
    ...
    if (!$scrollDownwards) {
        ...
        screenY + screenH + 24
        ...
        screenY + screenH + 46
        ...
    } else {
        ...
        ... screenY - 34
        ... screenY - 12
        ... screenY - 12

const drawTextLine ...
    ...
    ctx.fillRect(x - tw / 2 - padding, y - 17, tw + padding * 2, 22);
```

Current (Mac OS and Windows):
Image

After applying the suggested changes (Mac OS and Windows):
Image

simonwiles: I just committed `7f781649` before this comment was sent, which changes this up a bit.  Combine bits of both?

broadwell: OK yeah I'll give it a try again on my assorted devices...

src/components/RollViewer.svelte
```diff
@@ -695,30 +710,135 @@
...
-    marks = marks.filter(([hole, elem]) => {
-      if (holes.includes(hole)) return true;
-      viewport.viewer.removeOverlay(elem);
-      return false;
+    // Announce active holes
+    holes.forEach((hole) => {
+      announcement = hole.label.replace("#", "♯").replace("_", " ");
     });
 
+    const canvasRect = highlightCanvas.getBoundingClientRect();
```
broadwell: This variable appears to be unused. Of course it's unlikely to be the only case in the entire codebase -- actually I have a couple of cleanup changes I'll add to another PR eventually, but maybe we can fold this into the current one.

src/components/RollViewer.svelte
broadwell: All right so the background screen is gone now on Mac and Windows, which seems OK -- with the drop shadow, the white letters are still quite visible to me even against the most paperwhite-colored 88-note rolls.

And I'm happy with the distance at which the labels appear for Red Welte rolls (the only type for which they appear above the holes). To get the equivalent distance from the text top to the bottom of the holes for the rest of the rolls, I needed to use the following offsets:
screenY + screenH + ~~36~~ 28
and
screenY + screenH + ~~58~~ 50

If they look OK to you with those numbers on Linux browsers, I think we're good to go.

src/components/RollViewer.svelte
simonwiles: It's gone :D  91d9e7ef

src/components/RollViewer.svelte
```diff
@@ -695,30 +711,156 @@
...
+      const screenW = (holeW / imgBounds.width) * viewerSize.x;
+      const screenH = (holeH / imgBounds.height) * viewerSize.y;
+
+      // Get highlight color
+      let color = hole.color;
+      if (
+        !$userSettings.showNoteVelocities ||
+        $userSettings.highlightEnabledHoles
+      ) {
+        color = "60, 100%, 50%"; // yellow default for non-velocity mode
```
broadwell: Probably we should return to the previous behavior, which is that even with all fancy highlighting options turned off, the control holes are still colored green and the pedal holes colored orange.

Image

src/components/RollViewer.svelte
simonwiles: Done ✅️ 4be51f2e

src/components/RollViewer.svelte
simonwiles: Yeah, thanks, that's an oversight on my part.  Fixed in 4be51f2e.

broadwell (changes requested):
Another perhaps overly nuanced aspect of the visualization and emulation menu options: if "Display Note Velocities" is active and "Play Expressions" is not, the notes should all have velocity 64 (which is correct in this new branch), and the active note overlays should be colorized as if they have v=64, which is kind of a pale blue (this is perhaps not the interpretation of "Display" that some might expect, but it's what we've been going with for a while). The version in this branch currently colorizes them as if they still have the velocities they'd have if expression were enabled.

In this branch:
Image

Desired:
Image

simonwiles:
Ooh, that's a subtle one, good catch -- I'll see if I can get that fixed this morning.

Thanks for doing such a detailed review; some of this performance-orientated rewrite (specifically this canvas-based highlighting) has been in the back of my mind on-and-off for several years at this point, and you could make the case that now is perhaps not the best time to suddenly get the urge to implement it...  (I guess the best time was long ago, and the second best time is now, as they say?)  Your help in catching all these edge- and corner-cases is making me feel much more comfortable with it, though, and the difference in the user experience on anything but the best case is worth it, I think.

simonwiles:
Okay, I think `5f8dd44b` fixes that (last?) bit (it turned out to be a bit simpler than I first thought).

broadwell:
This all checks out for me now. Thanks for the persistence to get all of this in, and for volunteering the improvements in the first place!
Just making a note here to maybe revisit in a later branch the question of whether to bring back the display of the decimals in the velocity values when using in-app expression, rather than rounding them to integers. Unlike the expression MIDI generation code, which assumes a MIDI 1.0 instrument with only 7 bits for velocity values, the in-app expression feature of the app can and does calculate and play finer velocity gradations -- though whether most people can hear them is a different question.

PR #246, opened 2026-04-28
Misc updates to support screen readers and keyboard navigation, primarily

broadwell:
As the title indicates, the main goal here is to improve the reachability of various interface elements via keyboard navigation and to support more informative descriptions of their features by screen readers. Doing so does introduce some additional complexity to the app, for an unclear degree of true accessibility improvement.

### DIFF
--- a/src/components/ExpressionSettings.svelte
+++ b/src/components/ExpressionSettings.svelte
@@ -168,14 +168,14 @@
         <legend>Expression Settings</legend>
         {#each Object.keys($expressionParameters.tunable || {}) as expressionParam}
           <div>
-            <label class="exp-param" for={`"input_"${expressionParam}`}
+            <label class="exp-param" for={"input_" + expressionParam}
               >{$expressionParameters.tunable[expressionParam].alias}:</label
             >
             <div>
               <input
                 disabled={$appWaiting}
                 class="param-value"
-                id={`"input_"${expressionParam}`}
+                id={"input_" + expressionParam}
                 type="number"
                 min={$expressionParameters.tunable[expressionParam].min}
                 max={$expressionParameters.tunable[expressionParam].max}
--- a/src/components/ExpressionSettings.svelte
+++ b/src/components/ExpressionSettings.svelte
@@ -176,6 +176,8 @@
                 disabled={$appWaiting}
                 class="param-value"
                 id={`"input_"{expressionParam}`}
+                aria-label={$expressionParameters.tunable[expressionParam]
+                  .alias}
                 type="number"
                 min={$expressionParameters.tunable[expressionParam].min}
                 max={$expressionParameters.tunable[expressionParam].max}
--- a/src/components/ExpressionSettings.svelte
+++ b/src/components/ExpressionSettings.svelte
@@ -168,16 +168,14 @@
         <legend>Expression Settings</legend>
         {#each Object.keys($expressionParameters.tunable || {}) as expressionParam}
           <div>
-            <label class="exp-param" for={`"input_"{expressionParam}`}
+            <label class="exp-param" for={`"input_"${expressionParam}`}
               >{$expressionParameters.tunable[expressionParam].alias}:</label
             >
             <div>
               <input
                 disabled={$appWaiting}
                 class="param-value"
-                id={`"input_"{expressionParam}`}
-                aria-label={$expressionParameters.tunable[expressionParam]
-                  .alias}
+                id={`"input_"${expressionParam}`}
                 type="number"
                 min={$expressionParameters.tunable[expressionParam].min}
                 max={$expressionParameters.tunable[expressionParam].max}
--- a/src/components/RollViewer.svelte
+++ b/src/components/RollViewer.svelte
@@ -186,32 +186,42 @@
   let highlightCtx = null;
   const _highlightActivation = new WeakMap();
 
+  const getHoleDescription = (hole) => {
+    let holeLabel = hole.label;
+    let velocity = "";
+
+    // We only want to transpose notes, not ALL midi keys.
+    if (hole.type === "note") {
+      holeLabel = getHoleLabel(
+        hole.m + $transposeHalfStep,
+        $rollMetadata.ROLL_TYPE,
+      );
+
+      if ($userSettings.showNoteVelocities) {
+        velocity = Math.round(
+          $playExpressionsOnOff ? (hole.v ?? 64) : 64,
+        ).toString();
+      }
+    }
+
+    return [holeLabel, velocity];
+  };
+
   const createMark = (hole) => {
     const {
       x: offsetX,
       startY: offsetY,
       w: width,
       h: height,
-      m: midiKey,
-      v: velocity,
       color: holeColor,
       type: holeType,
     } = hole;
 
     const mark = document.createElement("mark");
 
-    // We only want to transpose notes, not ALL midi keys.
-    let transpose = 0;
-    if (holeType === "note") {
-      mark.dataset.noteVelocity = velocity || 64;
-      transpose = $transposeHalfStep;
-    }
-
-    const holeLabel = getHoleLabel(
-      midiKey + transpose,
-      $rollMetadata.ROLL_TYPE,
-    );
+    const [holeLabel, velocity] = getHoleDescription(hole);
     mark.dataset.holeLabel = holeLabel;
+    mark.dataset.noteVelocity = velocity || 64;
 
     mark.style.setProperty("--highlight-color", `hsl(${holeColor})`);
     mark.classList.add(holeType);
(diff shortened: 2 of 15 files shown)


src/components/RollViewer.svelte
```diff
@@ -636,14 +686,30 @@
       rect.setAttribute("height", height + padding * 2);
       rect.setAttribute("rx", 10);
       rect.setAttribute("ry", 10);
+      rect.setAttribute("fill", `hsla(${holeColor}, 0.8)`);
+      rect.setAttribute("class", holeType);
+      const [holeLabel, velocity] = getHoleDescription(hole);
+      const holeDescription = holeLabel
+        .replace("#", "-sharp")
+        .replace("_", " ")
+        .replace("A-sharp", "ae-sharp");
```
simonwiles: I don't think this is necessary at this point?  I'd vote to remove it, fwiw.

src/ui-components/SkipLink.svelte
```diff
@@ -27,4 +27,11 @@
   }
 </style>
 
-<a href="#content">Skip to main content</a>
+<script>
+  export let targetAnchor;
+  export let skipText;
+
+  console.log(targetAnchor, skipText);
```
simonwiles: Stray `console.log` :)

src/config/keyboard-shortcut-config.js
```diff
@@ -197,3 +197,10 @@ export const deltaControls = {
     help: "Amount to increase/decrease tempo by when holding the right-hand augment key and using the Tempo Up/Down keys",
   },
 };
+
+export const getShortcutKeyDesc = (keyMapping) =>
+  Object.values(alternativeIndicatorText).includes(keyMapping)
+    ? Object.keys(alternativeIndicatorText).find(
+        (key) => alternativeIndicatorText[key] === keyMapping,
+      )
+    : keyMapping;
```
simonwiles: This does an O(n) (double) lookup for every element on every render, I think?  Perhaps:
```js
const alternativeIndicatorRev = Object.fromEntries(
  Object.entries(alternativeIndicatorText).map(([k, v]) => [v, k])
);
export const getShortcutKeyDesc = (keyMapping) => alternativeIndicatorRev[keyMapping] ?? keyMapping;
```
or something like that?

src/components/ExpressionSettings.svelte
```diff
@@ -176,6 +176,8 @@
                 disabled={$appWaiting}
                 class="param-value"
                 id={`"input_"{expressionParam}`}
```
simonwiles: Not changed in this PR, but this is pretty broken...  It outputs `"input_"{expressionParam}` for every ID (likewise the `for` attrib on L171) meaning that all the labels and inputs have the same ID.

src/components/RollViewer.svelte
```diff
@@ -636,14 +686,30 @@
...
+        announcement = holeDescription;
         if (marks.map(([_hole]) => _hole).includes(hole)) return;
         viewport.viewer.removeOverlay(hoveredMark);
         hoveredMark = createMark(hole);
       });
-      rect.setAttribute("fill", `hsla(${holeColor}, 0.8)`);
-      rect.setAttribute("class", holeType);
+      rect.addEventListener("focusin", () => {
+        if (viewport.getZoom() < 1) adjustZoom("resetZoom");
+        const holeAriaLabel = `${holeDescription} ${velocity ? `velocity ${velocity || 64}` : ""}`;
```
simonwiles: Something wrong here: if the `velocity ?` test passes then `velocity` must be truthy, so `${velocity || 64}` is just `${velocity}`.  I think this should probably be:
```js
const holeAriaLabel = `${holeDescription} velocity ${velocity || 64}`;
```
?

src/components/ExpressionSettings.svelte
```diff
@@ -176,6 +176,8 @@
...
+                aria-label={$expressionParameters.tunable[expressionParam]
+                  .alias}
```
simonwiles: Fwiw, I think this should be redundant if the for/id issue above is fixed.  My understanding is that a) `aria-label` should supersede the visible `<label/>` for SRs if present; but that b) the visible `<label/>` should be the preferred mechanism and that `aria-label` should not be used in this situation if not needed (["The first rule of ARIA"](https://www.w3.org/TR/using-aria/#rule1) applies here, I think).

simonwiles (commented):
Cool, there's a lot of great stuff here, thanks!

I've made a couple of minor comments on specific bits (and I'm happy to go ahead and address those myself if you agree with my takes).

One other thing: there appears to be no visual indication of focussed holes still, so without SR enabled I'm just a bit lost hitting tab and wondering what's going on (unless I've missed something?).  I suggest a quick and simple:
```css
    &:focus {
      stroke: yellow;
      stroke-width: 15px;
    }
```
for `svg rect` in `hole-highlighting.scss` (again, happy to add this).

I'm not convinced about the zoom behaviour -- I know it's to prevent it taking forever to tab through if the roll's zoomed right out, but it feels a bit random to me.  I don't think I have a better idea at this time, though, I think I'm LGTM on that :)

Thanks for all this extra work!

src/components/RollViewer.svelte
simonwiles: Huh, thought I only had L695 selected here -- I meant I don't think the "A-sharp" -> "ae-sharp" is necessary, sorry.

edit: it looks like all the snippets are declining to indicate the actual line I was commenting on and showing extra context.  More GitHub smh I guess...

broadwell: No problem -- I'm fine with removing the "ae-sharp" line. It occurs to me that there are some pretty sharply diminishing returns with this kind of pronunciation whack-a-mole. For example, even my well-mannered screen reader still reads "Concerto in A Major" as "Concerto in Uh Major." Perhaps it's better just to leave such nuances to hypothetical models (which probably exist already) that are able to clock the musical context and adjust their pronunciations accordingly.

src/ui-components/SkipLink.svelte
broadwell: 🤦

src/components/RollViewer.svelte
broadwell: Good catch. The idea here was just to avoid reading out "velocity null" (or something like that) if it's a control hole that has no meaningful velocity value. But it occurs to me that the situation is complicated further when it's a note hole but "Play Expressions" has been unchecked. In that case, reading out the expression velocity value (which will already have been assigned to the hole) is undesirable, and reading out the default of 64 isn't really necessary either, so I'll probably go with
```javascript
const holeAriaLabel = `${holeDescription} ${$playExpressionsOnOff && velocity ? `velocity ${velocity}` : ""}`;
```
unless a more terse formulation comes to mind.

src/config/keyboard-shortcut-config.js
broadwell: Thanks, yes this is much better.

broadwell:
> One other thing: there appears to be no visual indication of focussed holes still, so without SR enabled I'm just a bit lost hitting tab and wondering what's going on (unless I've missed something?).

I had added some clunky JS code to `RollViewer` to strongarm the fill color of the hole to yellow on `focusin` (and to revert it on `focusout`), but it turns out that only works on Chrome on Mac OS, and not Firefox (or other browsers on Linux, presumably). In any case, adding a rule in SCSS to change the outline stroke on focus makes much more sense, so I'll drop the prior implementation in favor of that.

src/components/ExpressionSettings.svelte
broadwell: Whoops! Yes, both of those lines are missing a pretty important `$` -- and when it's added, the need for a separate `aria-label` attribute goes away.

src/ui-components/SkipLink.svelte
broadwell: Also, only tangentially related, but I might as well mention it here: the skip link to the pedal and accent controls was previously not wrapped in the necessary `{#if $appMode === "perform"}` check, so I've made that change in `Pianolatron.svelte`.

src/components/ExpressionSettings.svelte
broadwell: Indeed -- fixing the error you noticed re: the input field labels solves this problem.

src/components/ExpressionSettings.svelte
```diff
@@ -168,14 +168,14 @@
         <legend>Expression Settings</legend>
         {#each Object.keys($expressionParameters.tunable || {}) as expressionParam}
           <div>
-            <label class="exp-param" for={`"input_"{expressionParam}`}
+            <label class="exp-param" for={`"input_"${expressionParam}`}
```
simonwiles: This is still problematic, because now the id is (e.g.) `"input_"slow_decay_rate` (although at least they're unique per control now :)).

What we want is probably `{"input_" + expressionParam}` or ```{`input_${expressionParam}`}``` (I'd probably lean to the former.

simonwiles (changes requested):
Okay, great, with one small nit I think this is good to go now.  Thanks for getting the `holeAriaLabel` logic right 👍

src/components/ExpressionSettings.svelte
broadwell: Yeah let's go with the former. Sorry, I should have checked how the IDs actually looked in the DOM. I was somewhat confused by an apparent race condition in Firefox (or maybe between Firefox and the Mac OS VoiceOver utility) that causes glitches in how it reads out those stepper input fields, and led me to believe that the formatting above works better than others (which is not actually true).

simonwiles (approved):
Okay, let's get this in.  Thanks!

PR #247, opened 2026-05-11
Even more a11y

simonwiles:


src/components/RollViewer.svelte
```diff
@@ -909,7 +983,6 @@
     t < 0.5 ? 4 * t * t * t : 1 - Math.pow(-2 * t + 2, 3) / 2;
 
   // remove the current highlights and re-add them. Needed for when a transpose has taken place
-  // NOTE: This appears to be redundant?
```
simonwiles, line 912: I take it this is _not_ redundant, then?  I couldn't work out why it was necessary, nor could I determine any conditions under which it made any difference.  Not a big deal, just curious...

simonwiles (commented):
This is all good, sorry I've not got it in sooner.  I've created this PR first, and this review -- I'm going to merge this as-is, but I'm a bit puzzled about 6bdbbbe1 "Ensure holes receive keyboard focus in TTB+LTR order".  Under what circumstances is DOM order not automatically TTB+LTR?  I started to look at implementing this and decided it was unnecessary; what did I miss?

broadwell:
Re: https://github.com/sul-cidr/pianolatron/commit/6bdbbbe1d4a9436d323e3f82eb00f17a35bcfbdf - when the viewport is moving up the roll -- so when playing back a roll type that isn't red Welte, or when the user is manually scrolling upwards -- the newly visible SVG partitions will be added to the viewer overlay list in the DOM after/below those that were previously added and remain on screen. If the partitions aren't subsequently reordered, the tab-selection order will cycle first through the holes in the partitions that are now lower in the viewer, before jumping up to those in the top partition.

It's always possible that I misinterpreted the behavior I noticed before implementing the fix, but otherwise, I'm pretty sure the need is legitimate, even if the implementation is rather clunky.

src/components/RollViewer.svelte
broadwell: Yeah, this (or something like it) is needed if we want the hole labels of the currently highlighted holes (the ones still under the tracker bar) to change when a transposition is applied. It's most obvious when the playback is paused. Of course the actual notes aren't re-struck in this situation, so it does lead to a bit of a mismatch between the audio and the labels, but I think it's the more desirable behavior.

PR #248, opened 2026-05-12
Improve arrow-key navigation and item readouts from roll selector dropdown with JAWS screen reader

broadwell:
The title and commit messages pretty much say it all. The problems meant to be addressed here are only evident with Windows-based browsers when using JAWS.


simonwiles (approved):
Great, thanks!

PR #249, opened 2026-05-27
Streamline hole highlighting behavior

broadwell:
I noticed some behaviors related to hole highlighting that didn't seem right to me:

- expression hole highlighting being tied to the Use Roll Pedaling button rather than Play Expressions
- control and pedal holes flashing their base colors rather than yellow when active with Highlight Enabled Holes selected

so I set out to correct and somewhat streamline the hole highlighting code and the behavior of the Color Legend. Hopefully it all makes some kind of sense.


simonwiles (approved):
Yeah, this is all cool with me -- thanks!

PR #250, opened 2026-05-29
Embed search page has white background, like Spotlight exhibits

broadwell:
PRing this for now to get the preview link to add to the demo Spotlight exhibit... but probably it can just be merged?


simonwiles (approved):
Yeah, this makes a lot of sense.  My only thought is whether we might not just do the same with the regular search page, tbh.

Image

PR #251, opened 2026-06-02
Harmonize app design with proposed Spotlight exhibit changes

broadwell:
Requesting a review before the chain of commits gets too unwieldy (although that ship may already have sailed after the catalog record integration). Some of the style and layout updates probably could be done more elegantly, but as long as they work properly, I'm also fine with keeping them as is.


simonwiles:
Is the plan to request a review when it's ready for merging?

broadwell:
Yes -- thanks for checking. The PR should be coming soon. I don't anticipate adding any more new commits, but need to do some cleanup/rebasing.

simonwiles:
Okay, this all looks fine to me.  It took me a while to work out how to hide the header, tbh (and I knew the functionality was there and was looking for it), but I agree that the mousewheel solution had some drawbacks.  Merging now.

simonwiles (approved):

broadwell:
Thanks! Yeah the button is pretty hard to find if you don't know where to look. We might iterate on that in the future. But for now, it's great for this litany of commits to be merged.

PR #253, opened 2026-06-13
Use enharmonic pitch spellings in popups and voiceover when provided in roll JSON

broadwell:
Spellings estimated by the [pkspell](https://github.com/fosfrancesco/pkspell) model (credited).

Efforts were made to minimize the increase in the size of the per-roll JSON files, such as not storing elements of the note name that can be inferred from the MIDI number, like the octave, or the note name when there is no sharp or flat added.

No effort is made, however, to provide realistic pitch spellings when the transposition control is active. This is theoretically possible given that the pkspell model also estimates the underlying key signature, but that's beyond the scope of this project.

Also included: a couple of parasite commits involving roll metadata, as well as the Stravinsky Duo-Art rolls, which are currently accessioned with upside-down images but which I'm loathe to remove because they're our only examples of Duo-Art rolls at present.


simonwiles (approved):
This is very cool, really pleased it worked out.  I hope someone at SODA will be pleased :)

PR #254, opened 2026-06-15
Surprise fixes

broadwell:
The elaborated commit messages should help explain why these fixes are necessary. I'm happy to discuss them further, though.

**Update:** This should be ready for review now. I suppose it would be nice to know the root cause of the behavior https://github.com/sul-cidr/pianolatron/pull/254/commits/33cf2c826e4df9fcb490a195b0b27b457b9e4d2b addresses, but since it's likely buried in one of the dependencies, I can live with the not knowing, as long as we've banished the problem.

FYI, the bad behavior should be reproducible (pre-merge) by loading https://pianolatron-develop.netlify.app/perform/?druid=xn815zk7379, then skipping to beyond 70% and starting playback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@tonejs/midi": "^2.0.28",
     "@types/watch": "^1.0.6",
     "CanvasSpliner": "jonathanlurie/CanvasSpliner",
-    "midi-player-js": "^2.0.16",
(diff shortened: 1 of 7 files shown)


simonwiles (approved):
This all looks solid to me, and works as expected.  Merging now, thanks!

PR #258, opened 2026-06-22
Provide introductory tour via driver.js

broadwell:
This should be ready for review now. Some updates to unrelated UX elements found their way in, and have been flagged in the commit messages as such.


simonwiles:
Cool, the tour is a good addition to the site (and the other bits all check out for me), thanks for doing this.  If we do more on the tour it might be nice to move it out of the mail `Pianolatron.svelte` component, which is getting a bit lengthy now.

(oh, and there are a couple of odd bits of formatting in here that make me wonder if your Prettier setup is working properly? not a big deal, but unexpected.)

PR #260, opened 2026-06-29
Replace VelocitySpliner with a bunch of sliders, and some other things

broadwell:
The main purpose of this PR is to replace the very a11y-unfriendly VelocitySpliner widgets with some sliders that are much more keyboard- and screen-reader friendly, although sadly not as functional nor as cool looking.

Some secondary changes involve making the popup notifications and dialog modals more accessible and generally just a bit more usable, some follow-on tweaks to the layout, terminology and behaviors of the settings menus, and a couple of other loose ends.


src/components/Tour.svelte
```diff
@@ -0,0 +1,204 @@
...
+          description:
+            "These buttons display the current state of the pedals, just like the pedal graphics below the keyboard visualization, and also can be clicked to override the pedaling settings on the roll, as well as to boost (accent) the volume of playback temporarily.",
+        },
+      },
+      {
+        element: "button[aria-label='Performance controls']",
+        popover: {
+          title: "Performance Controls",
+          side: "top",
+          alignt: "start",
```
simonwiles, line 87: Pretty sure this should be `align` not `alignt` (although, since `"start"` is the default value, I don't think it makes any difference - https://driverjs.com/docs/configuration).

simonwiles (approved):
This all seems straightforward and consistent; looks good 👍️.  Thanks also for splitting out the `<Tour/>` component.

My only additional suggestion is perhaps the "Tour" link could get a little emphasis (an outline or underline or something) which could be removed if the tour has been started.  A different localStorage key could be used for play and perform modes, since there is more tour to be had in perform mode.  But maybe this isn't really needed 🤷.

